PR #478, opened 2017-02-28
Range slider

chrisdugne:
- `Handles` use hammerjs to listen events
- `RangeSlider` implements logic to move the 2 `Handles`


chrisdugne:
[(5.5) Double slider time](https://trello.com/c/H7BtThUZ/148-5-5-double-slider-time)

PR #506, opened 2017-03-15
Enhance range slider

chrisdugne:
- refacto to add `forceRange` feature
- closest step calculation as external util


chrisdugne:
[(5.5) Double slider time](https://trello.com/c/H7BtThUZ/148-5-5-double-slider-time)

PR #521, opened 2017-03-24
fix search component

esa-coorp:


esa-coorp:
@chrisdugne rebase done

PR #581, opened 2017-04-14
Separate freemium and premium components

tib-tib:


packages/@coorpacademy-components/src/molecule/tab-navigation/index.js
```diff
@@ -25,15 +25,15 @@ const TabNavigation = props => {
...
+        {map(tab => (
           <Tab
             {...otherProps}
             {...tab}
             onClick={onTabClick}
             key={tab.key}
             active={tab.key === activeContent}
           />
-        ))}
+        ), tabs)}
```
Bernardstanislas: 🤓

PR #638, opened 2017-05-10
fix filter animation on safari

esa-coorp:


packages/@coorpacademy-components/src/molecule/filters/style.css
```diff
@@ -139,7 +137,7 @@
 }
 
 .choice {
-  width: 22.5%;
+  width: 23%;
```
chrisdugne: utilise `box-sizing` plutot

PR #643, opened 2017-05-11
change color tabs

esa-coorp:


packages/@coorpacademy-components/src/variables/colors.css
```diff
@@ -2,6 +2,9 @@
 @value darkGrey: #484848;
 @value borderGrey: #d8d8d8;
 @value light: #F7F7F7;
+@value selectTab: #ECEFF1;
+@value unselectTab: #FAFAFA;
```
chrisdugne: https://projects.invisionapp.com/boards/BD36OFPSPM84H/

chrisdugne: cc @MelkMerle

PR #655, opened 2017-05-17
set slides-player progression as optional

chrisdugne:


chrisdugne:
https://github.com/CoorpAcademy/components/pull/658

PR #774, opened 2017-06-28
show card > .lockWrapper only if card is disabled

dawee:


MelkMerle:
https://coorpacademy.slack.com/archives/C0GN5V20M/p1498668321596465 Pas mergeable selon moi

PR #834, opened 2017-07-20
Removes animation of life.

godu:


chrisdugne:
[Popin end : game over](https://trello.com/c/5ynCWsfH/65-popin-end-game-over)

PR #917, opened 2017-08-24
Vimeo player events

chrisdugne:
using https://github.com/u-wave/react-vimeo


chrisdugne:
[Modifications UX/UI du player youtube-like](https://trello.com/c/DHDsyVnq/152-modifications-ux-ui-du-player-youtube-like)

PR #927, opened 2017-08-28
fix mobile/tablet view for video within popin-correction

chrisdugne:


chrisdugne:
[Modifications UX/UI du player youtube-like](https://trello.com/c/DHDsyVnq/152-modifications-ux-ui-du-player-youtube-like)

PR #976, opened 2017-09-18
Migrating services and engine config

chrisdugne:
- data.configs.entities.ref@version
- services requiring engine now pass it as parameter directly


chrisdugne:
https://github.com/CoorpAcademy/components/pull/977

PR #1010, opened 2017-10-11
Trigger build on component bump

godu:
https://docs.travis-ci.com/user/triggering-builds/


godu:
[Automatiser les bumps des components/app-player](https://trello.com/c/op8IofK5/136-automatiser-les-bumps-des-components-app-player)

PR #1023, opened 2017-10-15
flip clue

esa-coorp:


esa-coorp:
@godu 100%
`slides-player/index.js | 100% <100%> (+1.14%)`

PR #1031, opened 2017-10-19
few fixes for IE

chrisdugne:


chrisdugne:
[Retours Header app player](https://trello.com/c/LW762o71/395-retours-header-app-player)

PR #1042, opened 2017-10-24
fix cta popi-correction layout

chrisdugne:


chrisdugne:
[Learner : Popin de fin](https://trello.com/c/qHlbb0W6/330-learner-popin-de-fin)

PR #1045, opened 2017-10-24
Popin end fixes

jomaora:


jomaora:
[Learner : Popin de fin](https://trello.com/c/qHlbb0W6/330-learner-popin-de-fin)

PR #1113, opened 2017-12-05
Add color props to ChechIcon

GuillaumeRZ:


chrisdugne:
[Bouton Start XX Level qui se stretch verticalement](https://trello.com/c/HSOe7BmA/499-bouton-start-xx-level-qui-se-stretch-verticalement)

PR #1114, opened 2017-12-05
sync translations

chrisdugne:


chrisdugne:
[Learner Popin End post forums](https://trello.com/c/HrXKlAkg/336-learner-popin-end-post-forums)

PR #1144, opened 2018-01-16
migration UserAnswer

chrisdugne:


chrisdugne:
[[API-PROGRESSION] Correction du modèle AnswerAction](https://trello.com/c/4OL8yLhY/566-api-progression-correction-du-mod%C3%A8le-answeraction)

PR #1149, opened 2018-01-18
added learner create-initial-state

chrisdugne:


chrisdugne:
changement d'architecture de createProgression/createState https://github.com/CoorpAcademy/components/pull/1155

PR #1154, opened 2018-01-19
Add function to select initial content

chrisdugne:


chrisdugne:
[[adaptive] refacto compute-next-step](https://trello.com/c/HROOKrem/509-adaptive-refacto-compute-next-step)

PR #1163, opened 2018-01-25
added star source to selectRule

chrisdugne:
`selectRule` now accepts

```js
source: {
  type: 'slide',
  ref: '*'
}
```

### DIFF
--- a/packages/@coorpacademy-progression-engine/src/rule-engine/test/select-rule.js
+++ b/packages/@coorpacademy-progression-engine/src/rule-engine/test/select-rule.js
@@ -376,3 +376,47 @@ test("should select chapterRule with 'variables' scope", t => {
   const nocustomChapterRule = selectRule(slideScopedRules, nocustomState);
   t.is(nocustomChapterRule, null, "Don't support 'variable' scoped target with custom field");
 });
+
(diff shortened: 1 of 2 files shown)


packages/@coorpacademy-progression-engine/src/rule-engine/test/select-rule.js
```diff
@@ -376,3 +376,47 @@ test("should select chapterRule with 'variables' scope", t => {
   const nocustomChapterRule = selectRule(slideScopedRules, nocustomState);
   t.is(nocustomChapterRule, null, "Don't support 'variable' scoped target with custom field");
 });
+
+test("should apply a rule with source.ref:'*'", t => {
```
jfmengels: should select

packages/@coorpacademy-progression-engine/src/rule-engine/test/select-rule.js
```diff
@@ -376,3 +376,47 @@ test("should select chapterRule with 'variables' scope", t => {
...
+  const value = 111;
+  const baseSource = {
```
jfmengels: baseRule

PR #1167, opened 2018-01-26
bigger font size for popin correction

MelkMerle:


MelkMerle:
@chrisdugne @dawee ping

PR #1178, opened 2018-02-05
adaptive - popinEnd on fail mode

esa-coorp:


packages/@coorpacademy-app-player/src/store/view/state-to-props/popin-end.js
```diff
@@ -105,9 +106,12 @@ const summaryHeader = ({translate}, {dispatch}) => state => {
...
       () => ({
         title: translate('Ooops'),
-        subtitle: translate('You are out of lives!'),
+        subtitle:
+          lives === 0
+            ? translate('You are out of lives!')
+            : translate('Game over'),
         fail: true,
-        lives: get('state.lives')(progression),
+        lives: lives,
```
jfmengels: `lives: lives,` --> `lives,`

Issue #1176, opened 2018-02-02
progression-engine racing


chrisdugne:
```
export type User = {
  id: string,
  displayName: string
};
 
export type Team = Array<User>;

Action
  author: User | Array<User>,
```

```
progression: { 
  "_id": "1234",
   engine,
   slides,
  state: { 
    userId_1: { 
      status: 'ongoing | waiting | done',
       nextContent,
       isCorrect,
      step: int,
       score: string
    },

    userId_2: { 
      status: 'ongoing | waiting | done',
      nextContent,
       isCorrect,
      step: int,
       score: string
    } 
  }
}
```

chrisdugne:
createProgression + createState done

https://github.com/CoorpAcademy/components/pull/1226/commits/200e683c4f954e44c8401fd314b3f256ba974b8b

https://github.com/CoorpAcademy/components/pull/1197

PR #1197, opened 2018-02-15
progression -> createRacing

chrisdugne:
#1176

- added racing progression + racing types
- added `authors: Array<string>` to every action
- using `setupAction.payload` as `initialState`


chrisdugne:
see https://github.com/CoorpAcademy/components/pull/1226

PR #1223, opened 2018-03-15
allow hide element on header

jomaora:


jomaora:
[First-connection / Change Password page optimization](https://trello.com/c/rRsXsZOl/666-first-connection-change-password-page-optimization)

PR #1238, opened 2018-03-27
bestscore returns now stars and no more progression

chrisdugne:
findBestScore retrieves `{stars}` and no more `{progression}`


chrisdugne:
[New Interface : la next slide se charge super lentement](https://trello.com/c/aG7Z0Dki/697-new-interface-la-next-slide-se-charge-super-lentement)

PR #1239, opened 2018-03-28
fixed best score bailout for stars==0

chrisdugne:


chrisdugne:
[New Interface : la next slide se charge super lentement](https://trello.com/c/aG7Z0Dki/697-new-interface-la-next-slide-se-charge-super-lentement)

PR #1247, opened 2018-04-04
adding log on validate state reducer

jomaora:


packages/@coorpacademy-progression-engine/src/reducers/validate.js
```diff
@@ -16,11 +16,11 @@ export default function validate(config: Config): (State, Action) => void {
               '',
               'nextContent.type',
               state
-            )}) does not match the content (${getOr('', 'payload.content.ref', action)} - ${getOr(
+            )}) does not match the with the given ${action.type} action data: content.ref (${getOr(
```
jfmengels: `does not match the with the given` C'est pas très anglais.

Suggestion :

`does not match the content of the given ${action.type} action. State content: ${state.nextContent.ref} vs action (content ${action.payload.content.ref>}, nextContent  ${action.payload.nextContent.ref>})` (remet les get hein)

PR #1263, opened 2018-04-24
Cards partner redesign

dawee:


chrisdugne:
https://github.com/CoorpAcademy/components/pull/1273

PR #1330, opened 2018-06-21
added UserPreferences

chrisdugne:


chrisdugne:
[abonnement emails: refacto mooc](https://trello.com/c/6uzAPwKU/816-abonnement-emails-refacto-mooc)

PR #1332, opened 2018-06-22
IE background image

chrisdugne:
moved `:before` as `div`


chrisdugne:
[IE player BG + footer](https://trello.com/c/6q8YbSt6/817-ie-player-bg-footer)

PR #1340, opened 2018-08-03
[Totem] engine + redux + components

chrisdugne:
- added package app-racing
- added progression-engine racing
- racing templates/components (using react-motion)
- team-builder template
- racing admin template


chrisdugne:
#1386

PR #1366, opened 2018-10-04
[Totem] Refacto question race

chrisdugne:


chrisdugne:
[remove left/right animation between question/race](https://trello.com/c/gbNU3Z9D/945-remove-left-right-animation-between-question-race)

PR #1382, opened 2018-10-16
enhanced question media/image display

chrisdugne:


chrisdugne:
[Design slide question - taille de l'image](https://trello.com/c/3zLFw9Uu/806-design-slide-question-taille-de-limage)

PR #1391, opened 2018-11-19
Fix lives icon visible as it should not

VincentCATILLON:

The lives icon is visible whereas it shouldn't in case of not having lives (different than having lives at 0).

capture d ecran 2018-11-19 a 14 21 49

**Result and observation**

The lives icon are not displayed anymore.

capture d ecran 2018-11-19 a 14 20 44

- [ ]  **Breaking changes ?**
- [ ] **Extra lib ?** 

**How tested ?**
 - [ ] Manual testing
 - [x] Unit testing


VincentCATILLON:
@godu You mean something generated ? And why so ?

godu:
> @godu You mean something generated ? And why so ?

Ava allows you to reuse test logic. https://github.com/avajs/ava/blob/master/docs/01-writing-tests.md#reusing-test-logic-through-macros

In addition, i don't think splitting each test in own files is relevant.

VincentCATILLON:
@godu Please check again, I moved all tests in one file as it's relevant. IMHO using macro is just moving logic into another function for something very basic, increase complexity and decrease readability.

Before:

capture d ecran 2018-11-19 a 15 09 01

After:

capture d ecran 2018-11-19 a 15 09 11

PR #1394, opened 2018-11-20
Launch lerna

Volubyl:

Previously, not every packages have got the same scripts written in their package.json file. Though, it was not easy to manage them using lerna. This PR, add missing scripts.

**Result and observation**

1. Now we can use `lerna run build` or `lerna run test'`or even `lerna run eslint`to perform those action on each package. It is also possible to perform those action on a specific package using the `-- scope <package name>` flag. (eg : `lerna run test --scope @coorpacademy-component`).

2. This PR also clean the main `package.json` file since several action could be triggered using lerna.

Technical Debt :

Unfortunately, this PR also introduce technical debt : on each package `.eslintignore` and `.eslintrc` file were added. Those file are basicaly copy-paste from the original files located at the repo root and may contains irrelevant rules.  This should be clean up in a second time.

- [ ]  **Breaking changes ?**  

- [ ] **Extra lib ?** 

**How tested ?**
 - [X] Manual testing
 - [ ] Unit testing


package.json
```diff
@@ -1,22 +1,21 @@
...
+    "lerna": "lerna",
     "prepare": "lerna run prepare",
-    "preava": "npm run generate:fixtures",
-    "ava": "cross-env NODE_ENV=test ava",
     "build": "lerna run build",
+    "test": "lerna run test",
+    "eslint": "lerna run eslint",
+    "eslint:fix": "lerna run eslint --fix",
     "clean": "lerna run clean && lerna clean --yes",
+    "poststatic": "mkdirp static && cp -r packages/@coorpacademy-components/static static/components && cp -r packages/@coorpacademy-app-player/static static/app-player",
```
chrisdugne: si tu peux remettre ca sous `static` que ca saute aux yeux stp

packages/@coorpacademy-components/package.json
```diff
@@ -104,7 +109,16 @@
       "babel-polyfill",
       "../../test/helpers/throw-warning"
     ],
-    "babel": "inherit"
+    "babel": {
+      "presets": [
+        "@ava/stage-4",
+        "flow",
```
chrisdugne: pas besoin de flow pour les components react normalement

VincentCATILLON:
FYI, we need some additional PR on this one to handle:
- [ ] babel in global scope
- [ ] fixing codecov per package
- [ ] update readme

.travis.yml
```diff
@@ -15,16 +15,48 @@ cache:
...
+    if [ "$i" = "$PACKAGE_NAME" ]; then
+      echo -e "${CYAN} $PACKAGE_NAME has changed, bootstraping ${NC}";
+      yarn lerna bootstrap --scope $i  -—include-filtered-dependencies;
+  fi
+  done
+
+script: >-
+  for i in $(yarn lerna updated --loglevel=silent); do
+    if [ "$i" = "$PACKAGE_NAME" ]; then
+      echo -e "${CYAN} $PACKAGE_NAME has changed, tesing ${NC}";
```
VincentCATILLON: typo found: tesing => testing

.travis.yml
```diff
@@ -15,16 +15,49 @@ cache:
...
```
godu:
```suggestion
      echo -e "${CYAN} $PACKAGE_NAME has changed, testing ${NC}";
```

godu: on peut faire des suggestions de modification maintenant.

AdrieanKhisbe:
Belle découverte 1 an plus tard cette pr ....

PR #1396, opened 2018-11-21
update style and behavior of the select multiple component

jvaillant-coorp:

Within the backoffice (edit brand part) we would like a better UX for editing allowed `supported language` field (we currently use `select` component with multiple mode enabled)

**Result and observation**

Now we will use the 'select-multiple component', the component has been specially
updated to be integrated into the backoffice
- add ability to support theme
- using SVG icon instead of font
- new return value (array) set by the new multiple prop

**How tested ?**
 - [x] Manual testing
 - [x] Unit testing


packages/@coorpacademy-components/src/molecule/brand-form-group/index.js
```diff
@@ -19,58 +21,62 @@ import SetupSlider from '../setup-slider';
...
+    default:
+      return <InputText {...field} />;
+  }
+};
 
-  const buildField = (field, index) => {
-    const input = buildInput(field);
+const buildField = (field, index) => {
+  const input = buildInput(field);
+  const cn = classnames(style.field, {[style.withMultipleSelect]: field.type === 'selectMultiple'});
```
Fausto95: className instead of cn

PR #1403, opened 2018-11-26
Clean pakagesjson

Volubyl:

As we are considering each package individually, each package.json file should be considered the same and only contains relevant packages name and commands.

**Result and observation**

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [X] Manual testing
- [ ] Unit testing


packages/@coorpacademy-translate/package.json
```diff
@@ -47,14 +45,5 @@
...
   },
-  "ava": {
-    "files": [
-      "src/**/test/**/*.js"
-    ],
-    "require": [
-      "babel-register"
-    ],
-    "babel": "inherit"
-  },
```
VincentCATILLON: Be careful, this package is tested :)

PR #1427, opened 2018-12-24
composition middleware/reducer fetch progression

chrisdugne:

**Result and observation**

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [ ] Unit testing


chrisdugne:
[POC composition store](https://trello.com/c/8ebiiHip/1027-poc-composition-store)

VincentCATILLON:
Seen with you, we'll make it later, as we don't need it for MVP

chrisdugne:
--> note: here is the reducer composition maybe useful later

PR #1435, opened 2018-12-28
fix dependcies issues

Volubyl:

There were some dependencies issues and the yarn.lock was totally broken. Here is the solution

**Result and observation**

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [ ] Unit testing

### DIFF
--- a/packages/@coorpacademy-components/package.json
+++ b/packages/@coorpacademy-components/package.json
@@ -49,7 +49,6 @@
   "dependencies": {
     "@coorpacademy/nova-icons": "2.0.0",
     "autoprefixer": "^8.4.1",
-    "babel-plugin-transform-react-remove-prop-types": "0.4.19",
     "classnames": "^2.2.5",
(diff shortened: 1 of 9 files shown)


chrisdugne:
attention gros update des deps, faut sassurer que tout passe là @Volubyl

packages/@coorpacademy-components/package.json
```diff
@@ -49,6 +49,7 @@
   "dependencies": {
     "@coorpacademy/nova-icons": "2.0.0",
     "autoprefixer": "^8.4.1",
+    "babel-plugin-transform-react-remove-prop-types": "0.4.19",
```
jomaora: @Volubyl pourquoi il est deux fois ? Est-ce que tu pourrais checker les diffs de tes PR quand tu modifies des packages pour éviter ce type d'incohérences.

PR #1436, opened 2019-01-04
Remove fill attributes on native components

VincentCATILLON:

This PR removes fill attributes unsupported with the current version of `react-native-svg` that we can use (6.x) and which doesn't support `currentColor` and `inherit` values for `fill` attribute.

**Result and observation**

Nothing UI, but native components won't have fill attributes anymore (except if it's "none").

- [ ] **Breaking changes ?**
- [ ] **Extra lib ?**

**Testing Strategy**

- [x] Manual testing
- [ ] Unit testing


Fausto95:
GOOOOD GAME

PR #1442, opened 2019-01-21
Create progression

chrisdugne:

currently every store is initialized with an external progression

added createProgression action within redux workflow

**Result and observation**

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [x] Unit testing


packages/@coorpacademy-player-store/src/actions/api/progressions.js
```diff
@@ -23,29 +26,49 @@ import {
...
-
-export const PROGRESSION_FETCH_REQUEST: string = '@@progression/FETCH_REQUEST';
-export const PROGRESSION_FETCH_SUCCESS: string = '@@progression/FETCH_SUCCESS';
-export const PROGRESSION_FETCH_FAILURE: string = '@@progression/FETCH_FAILURE';
+import type {Dispatch, DispatchedAction, GetState, ThunkAction} from '../../definitions/redux';
 
-export type ProgressionAction = {
-  ...Action,
-  payload: Progression
+export type Action = {
```
VincentCATILLON: Can you use strict type ?

packages/@coorpacademy-player-store/src/actions/api/progressions.js
```diff
@@ -23,29 +26,49 @@ import {
...
+export const createProgression = (
+  engine: Engine,
+  content: Content,
+  engineOptions: EngineOptions
+): ThunkAction => async (
+  dispatch: Function,
+  getState: GetState,
+  {services}: {services: Services}
+): // $FlowFixMe circular declaration issue with gen-flow-files : type ThunkAction = (Dispatch, GetState, Options) => DispatchedAction
+DispatchedAction => {
```
VincentCATILLON: Actions should definitely return `Action` instead of dispatched ones, this is the role of middleware to do that. This is why you have circular issues. Every store property (actions+reducer) should live alone, and don't need to share any type. But at this time, this is ok if you understand the arguments of why it's not a good idea to do that in the future

PR #1447, opened 2019-01-22
truncate label in multi select overlay

esa-coorp:
before:
capture d ecran 2019-01-22 a 15 23 11

after:
capture d ecran 2019-01-22 a 15 22 46

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [x] Manual testing
- [ ] Unit testing


esa-coorp:
[Multiple select input form doesn't accept long string value](https://trello.com/c/HHeWQYYD/286-multiple-select-input-form-doesnt-accept-long-string-value)

PR #1471, opened 2019-01-30
Fix bug with computeNextSlide

Fausto95:

This PR was found in the mobile project. 
When we only have a slide and try to play, we get an error because there a 0 remaining slide, so if we try to get an empty _id the app crashes because there aren't any verification to avoid this 

**Result and observation**

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [ ] Unit testing

### DIFF
--- a/packages/@coorpacademy-progression-engine/src/compute-next-step/compute-next-step.js
+++ b/packages/@coorpacademy-progression-engine/src/compute-next-step/compute-next-step.js
@@ -165,9 +165,15 @@ const computeNextSlide = (
     pipe(get('_id'), (slideId: string) => !state || !includes(slideId, state.slides)),
     chapterContent.slides
   );
+  const nextSlide = pickNextSlide(remainingSlides);
+  if (isEmpty(nextSlide)) {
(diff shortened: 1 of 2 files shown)


packages/@coorpacademy-progression-engine/src/compute-next-step/compute-next-step.js
```diff
@@ -165,9 +165,15 @@ const computeNextSlide = (
     pipe(get('_id'), (slideId: string) => !state || !includes(slideId, state.slides)),
     chapterContent.slides
   );
+  const nextSlide = pickNextSlide(remainingSlides);
+  if (isEmpty(nextSlide)) {
```
chrisdugne: let's write a UT for this case

chrisdugne:
deprecated by https://github.com/CoorpAcademy/components/blob/master/packages/@coorpacademy-progression-engine/src/compute-next-step/compute-next-step.js#L94

PR #1485, opened 2019-02-12
export missing type

Volubyl:

**Result and observation**

Update types and export missing ones

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [X] Manual testing
- [ ] Unit testing

### DIFF
--- a/packages/@coorpacademy-player-store/src/definitions/redux.js
+++ b/packages/@coorpacademy-player-store/src/definitions/redux.js
@@ -44,12 +44,12 @@ type Data = {
 };
 
 type Ui = {
-  answers: Array<Answer>,
+  answers: {[key: string]: Answer},
   coaches: {
     availableCoaches: number
   },
   comments: {
-    text: string
+    text: ?string
   },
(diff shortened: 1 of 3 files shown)


packages/@coorpacademy-player-store/src/definitions/redux.js
```diff
@@ -44,12 +44,12 @@ type Data = {
 };
 
 type Ui = {
```
VincentCATILLON: comment: could be strict

packages/@coorpacademy-player-store/src/definitions/redux.js
```diff
@@ -44,12 +44,12 @@ type Data = {
...
-  answers: Array<Answer>,
+  answers: {[key: string]: Answer},
   coaches: {
     availableCoaches: number
   },
   comments: {
-    text: string
+    text: ?string
```
VincentCATILLON: why a maybetype ?

PR #1509, opened 2019-03-05
Fix type question slider

Fausto95:
Adding common question structure to slider question type

- [ ] Manual testing
- [ ] Unit testing


packages/@coorpacademy-progression-engine/src/types.js
```diff
@@ -272,6 +272,7 @@ export type QCMGraphicQuestion = $Exact<{|
 |}>;
 
 export type SliderQuestion = {|
+  ...QuestionCommon,
```
VincentCATILLON: It could be kind to have this update in `BasicQuestion` too

Fausto95: Done

PR #1517, opened 2019-03-13
Adding settings target icons

alanlanglois:
Add icons settings and targer, removing styles in existing svg of these two icons. 

**Result and observation**

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [ ] Unit testing


Fausto95:
GG🎉

PR #1531, opened 2019-03-22
Exporting new icon

Fausto95:
Exporting an existing `(X)` icon to the lib

- [ ] Manual testing
- [ ] Unit testing


Fausto95:
Exporting both now

PR #1532, opened 2019-03-22
fix english gramar into permission popin

marcmdm:
j'ai retiré le 1er "to" dans la phrase "We need **to** access to your camera in order to scan your QR code and authenticate you." sur conseil de faust'

### DIFF
--- a/packages/@coorpacademy-player-web/locales/en/player.json
+++ b/packages/@coorpacademy-player-web/locales/en/player.json
@@ -32,7 +32,7 @@
   "Open settings": "Open settings",
   "Ooops": "Ooops...",
   "Permission": "Permission",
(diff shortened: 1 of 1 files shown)


packages/@coorpacademy-player-web/locales/en/player.json
```diff
@@ -32,7 +32,7 @@
   "Open settings": "Open settings",
   "Ooops": "Ooops...",
   "Permission": "Permission",
-  "Permission camera": "We need to access to your camera in order to scan your QR code and authenticate you.",
+  "Permission camera": "We need access to your camera in order to scan your QR code and authenticate you.",
```
MelkMerle:
```suggestion
  "Permission camera": "We need to access your camera in order to scan your QR code and authenticate you.",
```

PR #1548, opened 2019-04-04
Uprade storybook version (and mostly its required dependencies)  :rocket:

Volubyl:

The main task was to upgrade storybook version from 4.x.x to 5.x.x but in order to achieve this some steps were required ...

So, instead of only of updatting storybook, those packages were updated :

- ava 0.x.x -> 1.x.x
- babel 6.x.x -> 7.x.X
- storybook  3.x.x -> 5.x.x

Thanks to @AdrieanKhisbe for having kickstarted that huge work 

**Result and observation**

Let me introduce ava 1.x.x and storybook 5.x.x 🎸 

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [X] Already covered by tests
- [X] Manual testing
- [X] Unit testing


packages/@coorpacademy-webpack-config/package.json
```diff
@@ -13,12 +13,13 @@
     "nyc": "cross-env NODE_ENV=test nyc npm run ava",
     "test": "npm run lint",
     "lint": "npm run eslint",
-    "eslint": "eslint ./ *.js *.json --ext .js,.json",
+    "eslint": "eslint **/*.js *.json --ext .js,.json",
```
godu: Nope. Remplaces par ça

```
eslint . --ext .js,.json
```

PR #1553, opened 2019-04-09
fix/package-version

Volubyl:

Fix packages version according to the most recent major release available on npm

**Result and observation**

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [ ] Unit testing


AdrieanKhisbe:
????
Why no reviews??

PR #1565, opened 2019-05-02
split progression updated onMove/onNode

chrisdugne:
**Detailed purpose of the PR**

```
- quand tu regardes un media sur la dernière popin interquestion du niveau (j’ai donc déjà gagné ou perdu), cela re-trigger un nouveau finish progression à chaque fois que je regarde un media
```

--> split `progressionUpdated` for node and move changes;
then 'finishProgression' will be tested `onMove` only

**Testing Strategy**

- [x] Already covered by tests
- [x] Manual testing
- [x] Unit testing

PR #1575, opened 2019-05-07
set sendProgressionUpdated optional

chrisdugne:
`sendProgressionUpdated` is not always (well, actually never as of today) implemented service-side


chrisdugne:
[Retours GTM](https://trello.com/c/4poy0K66/1279-retours-gtm)

chrisdugne:
simple oversight for https://github.com/CoorpAcademy/components/pull/1565

PR #1579, opened 2019-05-10
Change cursor on hover store images

Fausto95:
# Before 

not-hover

# After 

hover


Fausto95:
seen with @esa-coorp already!

PR #1588, opened 2019-06-11
Fix/fix data layer // allow find content to return void

Volubyl:

Fix/fix data layer // allow find content to return void

**Result and observation**

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [ ] Unit testing

### DIFF
--- a/packages/@coorpacademy-player-web/package.json
+++ b/packages/@coorpacademy-player-web/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@coorpacademy/player-web",
-  "version": "3.1.16-alpha.0+4c1ebf348",
+  "version": "3.1.15",
   "description": "CoorpAcademy web slide-player",
   "main": "lib/index.js",
   "module": "es/index.js",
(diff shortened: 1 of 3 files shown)


packages/@coorpacademy-player-services/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@coorpacademy/player-services",
-  "version": "5.0.6",
+  "version": "5.0.7-alpha.0+4c1ebf348",
```
chrisdugne: attention ne change pas ça

packages/@coorpacademy-player-web/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@coorpacademy/player-web",
-  "version": "3.1.15",
+  "version": "3.1.16-alpha.0+4c1ebf348",
```
chrisdugne: ici non plus

PR #1591, opened 2019-07-01
translation for video loading error

Fausto95:
Translation for mobile when trying to load a video with error


packages/@coorpacademy-player-web/locales/en/player.json
```diff
@@ -60,6 +60,7 @@
   "Upgrade description": "Your application has been updated! Please download the latest version to enjoy it.",
   "Validate": "Validate",
   "Version": "Version",
+  "Video Loading Error": "An error occurred while loading the video, please try again.",
```
VincentCATILLON: IMO this translation should be in components instead of player-web as it's not related to player-web

PR #1594, opened 2019-07-05
Adding state extract for videos

Fausto95:
This PR adds a state-extract function for 
```
state: {
   data: {
     contents: {
        video: {
           entities: {
               id: string
           }
         }
      }
  }
```

### DIFF
--- a/packages/@coorpacademy-player-store/src/utils/state-extract.js
+++ b/packages/@coorpacademy-player-store/src/utils/state-extract.js
@@ -493,3 +493,11 @@ export const hasSeenLesson = (state: State, onPreviousSlide: boolean = false): b
     isEmpty(lessons) || any(ref => includes(ref, map('ref', lessons)), viewedResourcesForContent)
   );
 };
+
+export const getVideoUrl = (id: string) => (state: State): string =>
+  state && // eslint-disable-line lodash-fp/prefer-get
+  state.data &&
+  state.data.contents &&
+  state.data.contents.video &&
+  state.data.contents.video.entities &&
+  state.data.contents.video.entities[id];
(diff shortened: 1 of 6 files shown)


packages/@coorpacademy-player-store/src/utils/state-extract.js
```diff
@@ -493,3 +493,11 @@ export const hasSeenLesson = (state: State, onPreviousSlide: boolean = false): b
...
   );
 };
+
+export const getVideoUrl = (id: string) => (state: State): string =>
+  state && // eslint-disable-line lodash-fp/prefer-get
+  state.data &&
+  state.data.contents &&
+  state.data.contents.video &&
+  state.data.contents.video.entities &&
+  state.data.contents.video.entities[id];
```
VincentCATILLON: it seems that it's not in `state.data.contents` but in `state.data` directly:

https://github.com/CoorpAcademy/components/pull/1594/files#diff-df77b9ccae3ea57d6dc59267aae00b5fR48

packages/@coorpacademy-player-store/src/utils/test/state-extract.js
```diff
@@ -866,6 +867,19 @@ test('getNbSlides should return 0 if no contentInfo is found', t => {
   return t.is(getNbSlides(state), 0);
 });
 
+test('getVideoUrl should return the video url', t => {
+  const videoId = 'DE3942dz9ke';
+  const state = set(
+    'data.contents.video.entities',
```
VincentCATILLON: same than above about the path (`data.contents.video -> `data.video`)

packages/@coorpacademy-player-store/src/utils/state-extract.js
chrisdugne: no, it's ok we'll use `fetchContent({type:video})`

VincentCATILLON: yes but the type is incorrect so (see the link above). the video state is not in contents state

VincentCATILLON: oh ok it was fixed after my first comment :) my bad

PR #1596, opened 2019-07-11
Make the videos state independent of content one

VincentCATILLON:

After analysis, videos state should not be related to content because it's not a content (content: slide, chapter, level, and sometimes a course, but not in player responsible area), but an independent state (neither an entire resource because we are not storing the entire resource, only the url)

- [ ] **Breaking changes ?**
- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [x] Unit testing


packages/@coorpacademy-player-store/src/actions/api/videos.js
```diff
@@ -0,0 +1,36 @@
...
+  {services}: {services: Services}
+): // $FlowFixMe circular declaration issue with gen-flow-files : type ThunkAction = (Dispatch, GetState, Options) => DispatchedAction
+DispatchedAction => {
+  const {Videos: VideosService} = services;
+
+  const action: Action = buildTask({
+    types: [VIDEOS_FETCH_URI_REQUEST, VIDEOS_FETCH_URI_SUCCESS, VIDEOS_FETCH_URI_FAILURE],
+    task: () => VideosService.findUriById(id, provider),
+    meta: {id, provider},
+    bailout: getVideoUri(id)
```
chrisdugne: 👍

PR #1603, opened 2019-07-16
added getNextChapter

chrisdugne:
microlearning was not handled by `RecommendationsService.getNext`.
added this behaviour to match mooc behaviour: https://github.com/CoorpAcademy/coorpacademy/blob/master/core/public/js/services/ContentService.js#L46-L53


chrisdugne:
[As a mobile user, when i have finished a level i can't access the next level.](https://trello.com/c/oqepKL6s/1370-as-a-mobile-user-when-i-have-finished-a-level-i-cant-access-the-next-level)

PR #1604, opened 2019-07-17
fix component battle-requests

rocket-star:

**Result and observation**

**Problème de base:** Les boutons du composant CardWithButton ne supportait pas toutes les traductions
**Fix:** modification du style pour permettre le support de toutes les traductions.

**BEFORE**
image

**AFTER**
image


VincentCATILLON:
I'm not sure to understand why changing some styles will change the languages support ?

Can you please add a before/after to see the difference and explain what this PR is is supposing to change ?

PR #1611, opened 2019-08-07
Handle JWP 214000 Error

Fausto95:
Some of our IE11 users on W7 are facing some errors while trying to play some videos due to a weak support of M3U8 video formats on IE11, so, to bypass this error here we're switch from M3U8 (aka HLS) video format to MP4 in order to let the user keep his experience on the platform without any pain

### DIFF
--- a/packages/@coorpacademy-components/src/molecule/video-player/jwplayer.js
+++ b/packages/@coorpacademy-components/src/molecule/video-player/jwplayer.js
@@ -65,7 +65,10 @@ class JWPlayer extends React.Component {
     // Since IE11 dont kinda support M3U8 sometimes,
     // We've decided to switch from M3U8 to mp4 whenever it appears
     if (code === 214000) {
-      const {videoId} = this.props.jwpOptions;
+      const {file: videoUrl} = this.props.jwpOptions;
+      const regex = /^https:\/\/content.jwplatform\.com\/manifests\/(\w+).m3u8/;
+      const videoId = videoUrl.match(regex)[1];
+
       this.setState({fileUrl: `https://content.jwplatform.com/videos/${videoId}-1080.mp4`});
     }
   }
@@ -89,7 +92,6 @@ class JWPlayer extends React.Component {
 JWPlayer.propTypes = {
   jwpOptions: PropTypes.shape({
     file: SrcPropType,
-    videoId: PropTypes.string,
     customProps: PropTypes.shape({
       aspectratio: PropTypes.string,
       tracks: PropTypes.shape({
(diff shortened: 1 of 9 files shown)


Fausto95:
[Bug : JWP errors 214000](https://trello.com/c/qrV5mjjJ/1383-bug-jwp-errors-214000)

packages/@coorpacademy-components/src/molecule/video-player/jwplayer.js
```diff
@@ -49,7 +79,8 @@ class JWPlayer extends React.Component {
         onResume={this.handleResume}
         onPause={this.handlePause}
         onOneHundredPercent={this.handleEnded}
-        {...this.props.jwpOptions}
+        onError={this.handleError}
+        {...{...this.props.jwpOptions, file: this.state.fileUrl}}
```
VincentCATILLON: maybe you can spread first in props, and define the file prop=this.state.fileUrl

packages/@coorpacademy-components/src/molecule/video-player/jwplayer.js
```diff
@@ -9,10 +9,18 @@ import style from './jwplayer.css';
...
+    };
     this.handlePlay = this.handlePlay.bind(this);
     this.handleResume = this.handleResume.bind(this);
     this.handlePause = this.handlePause.bind(this);
     this.handleEnded = this.handleEnded.bind(this);
+    this.handleError = this.handleError.bind(this);
+  }
+
+  componentDidMount() {
+    this.setFileUrl();
```
VincentCATILLON: this thing should be done in didUpdate or didReceiveProps to ensure that the url can change even if the component doesn't unmount

packages/@coorpacademy-components/src/molecule/video-player/jwplayer.js
```diff
@@ -9,10 +9,18 @@ import style from './jwplayer.css';
 class JWPlayer extends React.Component {
   constructor(props, context) {
     super(props, context);
+    this.state = {
+      fileUrl: ''
```
VincentCATILLON: comment: this state property can have the same name than the jwpOptions

packages/@coorpacademy-components/src/molecule/video-player/jwplayer.js
```diff
@@ -41,15 +58,32 @@ class JWPlayer extends React.Component {
...
+  handleError(error) {
+    const {code} = error;
+
+    // This error mostly appears on IE11 on W7
+    // Since IE11 dont kinda support M3U8 sometimes,
+    // We've decided to switch from M3U8 to mp4 whenever it appears
+    if (code === 214000) {
+      const {file: videoUrl} = this.props.jwpOptions;
+      const regex = /^https:\/\/content.jwplatform\.com\/manifests\/(\w+).m3u8/;
+      const videoId = videoUrl.match(regex)[1];
```
godu: `String.prototype.match`peut retourner `null`...

PR #1617, opened 2019-09-02
for adaptive course, don't show lifes in popinend 

esa-coorp:
Before:
Capture d’écran 2019-09-02 à 09 54 55
after:
Capture d’écran 2019-09-02 à 09 55 32

**Result and observation**

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [x] Manual testing
- [x] Unit testing

### DIFF
--- a/packages/@coorpacademy-components/src/molecule/feedback/index.js
+++ b/packages/@coorpacademy-components/src/molecule/feedback/index.js
@@ -1,6 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import getOr from 'lodash/fp/getOr';
+import pipe from 'lodash/fp/pipe';
+import get from 'lodash/fp/get';
+import extend from 'lodash/fp/extend';
+
 import ResourcePlayer, {TYPE_IMAGE, TYPE_VIDEO, TYPE_PDF} from '../resource-player';
 import style from './style.css';
 
@@ -9,7 +12,7 @@ const Feedback = (props, context) => {
   const resource = media &&
     media.type && {
       ...media,
-      ...getOr({}, 'src.0', media)
+      ...pipe(get('src.0'), extend({description: mediaDescription}))(media)
     };
 
   return (
--- a/packages/@coorpacademy-player-store/src/utils/state-extract.js
+++ b/packages/@coorpacademy-player-store/src/utils/state-extract.js
@@ -1,6 +1,7 @@
 // @flow strict
 
 import get from 'lodash/fp/get';
+import last from 'lodash/fp/last';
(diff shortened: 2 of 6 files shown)


packages/@coorpacademy-components/src/molecule/feedback/index.js
```diff
@@ -9,7 +12,7 @@ const Feedback = (props, context) => {
   const resource = media &&
     media.type && {
       ...media,
-      ...getOr({}, 'src.0', media)
+      ...pipe(get('src.0'), extend({description: mediaDescription}))(media)
```
VincentCATILLON: why do you need to change this logic whereas it's only to remove lives in popin-end header ?

packages/@coorpacademy-player-store/src/utils/state-extract.js
```diff
@@ -237,16 +237,30 @@ export const getPrevStepContent = (state: State): Content | void => {
...
+  if (progression.state.nextContent.type === 'slide') {
+    return getChapterIdBySlide(state, progression.state.nextContent.ref);
+  } else {
+    if (!progression.state.slides || isEmpty(progression.state.slides)) {
+      return;
+    }
+    const lastSlideId = last(progression.state.slides);
+    return getChapterIdBySlide(state, lastSlideId);
+  }
+};
```
VincentCATILLON: same for this

godu: => https://github.com/CoorpAcademy/components/pull/1617/files#diff-66e631ed454627bf02a2067a526bf492R803-R810
C'est dans le cas où on est sur un extraLife, le nextContent n'est plus la slide.

esa-coorp: `popinEnd` uses the `getLives` function to determine whether to show/hide life

=> https://github.com/CoorpAcademy/components/blob/master/packages/%40coorpacademy-player-web/src/map-state-to-props/popin-end.js#L79

=> https://github.com/CoorpAcademy/components/blob/master/packages/%40coorpacademy-player-store/src/utils/state-extract.js#L460

packages/@coorpacademy-components/src/molecule/feedback/index.js
esa-coorp: for a popinEnd with pdf ressource, the description of the resource was not displayed

packages/@coorpacademy-player-store/src/utils/state-extract.js
VincentCATILLON: ok 👍

PR #1627, opened 2019-09-19
Fix incorrect type in state extract (correction)

VincentCATILLON:

This PR fixes incorrect returned type because of lodash usage.

**Result and observation**

At this time, we got some errors on mobile app because of the incorrect type returned.

- [ ] **Breaking changes ?**
- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [x] Unit testing


packages/@coorpacademy-player-store/src/utils/test/state-extract.js
```diff
@@ -499,15 +505,15 @@ test('getCurrentCorrection should get current correction from state', t => {
   t.is(getCurrentCorrection(state), correction);
 });
 
-test('getCurrentCorrection should return "" if no progression is found', t => {
+test('getCurrentCorrection should return default correction if no progression is found', t => {
```
Fausto95: defaultCorrection

PR #1630, opened 2019-09-27
Fix findBestOf function

VincentCATILLON:

This PR changes the `findBestOf` and `getAllProgressions` functions need to be used as a promise because in mobile environment, we're dealing with promises for these ones.

**Result and observation**

Everything should be green and on `popin-end`, we should subtract the stars to the current best score.

- [ ] **Breaking changes ?**
- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [x] Unit testing

### DIFF
--- a/packages/@coorpacademy-player-services/src/progressions.js
+++ b/packages/@coorpacademy-player-services/src/progressions.js
@@ -48,7 +48,7 @@ type FindBestOf = (
   contentType: ContentType,
   contentRef: string,
   progressionId: string
-) => Progression;
+) => Promise<{|stars: number|}>;
 type FindById = (id: string) => Promise<Progression | void>;
 type GetAvailableContent = Content => Promise<AvailableContent>;
 type MarkResourceAsViewed = (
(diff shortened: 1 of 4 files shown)


packages/@coorpacademy-player-services/src/progressions.js
```diff
@@ -141,20 +141,23 @@ const createSave = (dataLayer: DataLayer): (Progression => Promise<Progression>)
...
   const {getAllProgressions} = dataLayer;
-  const progressions = getAllProgressions();
+  const progressions = await getAllProgressions();
 
-  const bestProgression = pipe(
+  const bestProgression: Progression | void = pipe(
     filter(p => get('content.ref', p) === contentRef && get('_id', p) !== progressionId),
     maxBy(p => (p.state && p.state.stars) || 0)
   )(progressions);
-  return bestProgression || set('state.stars', 0, {});
```
VincentCATILLON: this returned type was wrong (Progression | CustomObjectRefleatingProgression), and the reducer was expecting an object with `stars` property instead of `state.stars`

Fausto95:
LGTM, but will check more in depth Monday

PR #1631, opened 2019-10-04
Add ability to force best score refresh

VincentCATILLON:

This PR adds ability to force best score fetch to refresh it on mobile.

**Result and observation**

You should be able to refresh the best score even it's already fetched. It's useful if you have a store persisted between your app screens and needs to refresh this value to take in consideration local progressions or best score change on API.

- [ ] **Breaking changes ?** 
- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [x] Unit testing

### DIFF
--- a/packages/@coorpacademy-player-store/src/actions/api/progressions.js
+++ b/packages/@coorpacademy-player-store/src/actions/api/progressions.js
@@ -262,7 +262,8 @@ export const PROGRESSION_FETCH_BESTOF_FAILURE: string = '@@progression/FETCH_BES
 
 export const fetchBestProgression = (
   progressionContent: Content,
-  progressionId: string
+  progressionId: string,
+  force?: boolean
 ): ThunkAction => (
   dispatch: Function,
   getState: GetState,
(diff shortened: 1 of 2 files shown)


packages/@coorpacademy-player-store/src/actions/api/progressions.js
```diff
@@ -287,10 +288,12 @@ DispatchedAction => {
       PROGRESSION_FETCH_BESTOF_FAILURE
     ],
     task: () => Progressions.findBestOf(engine.ref, type, ref, progressionId),
-    bailout: pipe(
-      getBestScore,
-      score => score !== undefined && score >= 0
-    ),
+    bailout: force
```
VincentCATILLON: `bailout` is used to prevent requests if the data is already there, but in case of `bestScore`, we may need to refresh it to be sure to have the higher best score value and prevent displaying a `+xx stars` every time the  user finishes the same course without reloading the app to clean the state.

VincentCATILLON:
will be done in mobile

PR #1638, opened 2019-10-31
Dependencies and lint refresh

AdrieanKhisbe:
Basicly:
- update most dependencies (but react,angular and flow)
- set up new lint config version (fixes: mainly removing useless disable directives)
- lerna upgrad
- package.json clean up


AdrieanKhisbe:
(just rebased)

PR #1639, opened 2019-11-02
fix accordion style for setup

esa-coorp:
Fix accordion design on setup.
 - Add padding into slide of accordion
 - Add a border
 - Add tests

______________________ Before ______________________ ______________________ After ______________________:
Capture d’écran 2019-11-02 à 14 49 25Capture d’écran 2019-11-02 à 14 48 08

**Result and observation**

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [x] Manual testing
- [x] Unit testing


packages/@coorpacademy-components/src/organism/accordion/toggler/index.js
```diff
@@ -33,7 +33,7 @@ class AccordionToggler extends React.Component {
   render() {
     const {children = []} = this.props;
     return (
-      <Accordion tabProps={this.state.tabProps} onClick={this.handleOnClick}>
+      <Accordion tabProps={this.state.tabProps} onClick={this.handleOnClick} theme={'setup'}>
```
chrisdugne: c'est plutot à l'app de passer le theme en props non ?
je remonterais encore d'un cran je crois

PR #1649, opened 2019-11-13
new tab design

esa-coorp:
## OLD 
old-desktop

## NEW
desktop

**Result and observation**

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [ ] Unit testing


packages/@coorpacademy-components/src/molecule/module-bubble/style.css
```diff
@@ -45,6 +45,7 @@
   width: 38px;
   height: 38px;
   transition-duration: 0.5s;
+  color: dark;
```
VincentCATILLON: 👌

PR #1651, opened 2019-11-15
add sticky block on discipline page

esa-coorp:
### 1 -  desktop with sticky block
desktop-discipline
### 2 -  tablette without sticky block
tablette-description
### 3 -  mobile without sticky block

**Result and observation**

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [ ] Unit testing

### DIFF
--- a/packages/@coorpacademy-components/src/template/common/discipline/style.css
+++ b/packages/@coorpacademy-components/src/template/common/discipline/style.css
@@ -14,6 +14,14 @@
   margin-bottom: 30px;
 }
 
+.sticky {
+  position: fixed;
+  left: 50%;
+  margin-left: 330px;
+  top: 80px;
+  width: 280px;
+}
+
 .leftColumn {
   composes: column;
   float: left;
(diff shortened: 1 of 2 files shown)


packages/@coorpacademy-components/src/template/common/discipline/style.css
```diff
@@ -14,6 +14,14 @@
   margin-bottom: 30px;
 }
 
+.sticky {
+  position: fixed;
+  left: 50%;
+  margin-left: 330px;
```
VincentCATILLON: what about a container bigger in desktop ? if we have multiple container widths (ie: 1024, 1220, etc), we can have glitches with this sticky block overriding the content on the left. if not, it's ok.

comment only: maybe we could calculate this value to have something more "flexible" if we change the container width

esa-coorp: le container parent est en `margin: auto` donc centré et en `max-width 1280px` et le block de gauche est en `float: left;` et `width: 840px;` et cette règle ne s'applique que sur les écrans au dessus de 1280px de large.

PR #1655, opened 2019-12-03
[FIX]Trigger on play even with auto start

Volubyl:

Consider a resourced as viewed even after a video autostart.

**Result and observation**

Before

Kapture 2019-12-03 at 15 53 07

After

Kapture 2019-12-03 at 15 58 06

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [X] Manual testing
- [ ] Unit testing

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "flow": "lerna run flow",
     "codecov": "codecov",
     "publish:beta": "lerna publish --npm-tag beta --exact",
-    "publish:canary": "lerna publish --canary --exact patch --no-git-tag-version --no-push --npm-client npm",
+    "publish:canary": "lerna publish --canary --exact patch --no-git-tag-version --no-push",
     "publish:latest": "lerna publish --exact",
     "postinstall": "npm run preflow",
     "update": "update-node"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -13,7 +13,7 @@
     "flow": "lerna run flow",
     "codecov": "codecov",
     "publish:beta": "lerna publish --npm-tag beta --exact",
-    "publish:canary": "lerna publish --canary --exact patch --no-git-tag-version --no-push --npm-client npm",
+    "publish:canary": "lerna publish --canary --exact patch --no-git-tag-version --no-push",
```
Fausto95: What's that?

VincentCATILLON: I agree, it should be done in another PR.

FYI: https://github.com/lerna/lerna/blob/master/commands/publish/CHANGELOG.md#372-2018-12-21

Volubyl: If we keep those flags the script pops an error :

Screenshot 2019-12-03 at 16 25 18

VincentCATILLON: Introduced there: https://github.com/CoorpAcademy/components/pull/1638

VincentCATILLON:
I can't approve, because I worked on it, even if I'm ok

PR #1663, opened 2020-01-02
Just Yarn Bootstrap

Fausto95:
I dunno what happened during these days, but looks link the lint ain't happy.
These changes were created just by running `yarn bootstrap` in the root directory


jomaora:
@Fausto95 t'as publié une canary pour voir ce que ça donne sur les icons ?

PR #1669, opened 2020-01-09
Fix button click

Fausto95:
Use a cb function instead

### DIFF
--- a/packages/@coorpacademy-components/src/molecule/modal/index.js
+++ b/packages/@coorpacademy-components/src/molecule/modal/index.js
@@ -13,9 +13,9 @@ import PropTypes from 'prop-types';
 import Provider from '../../atom/provider';
 import style from './style.css';
 
-const Choice = ({title, subtitle, onClick, children}) => {
+const Choice = ({title, subtitle, handleOnClick, children}) => {
   return (
-    <a className={style.choice} onClick={onClick}>
+    <a className={style.choice} onClick={handleOnClick}>
       <div className={style.choiceContent}>
         <div className={style.choiceLeftIconWrapper}>{children}</div>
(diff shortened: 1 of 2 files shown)


packages/@coorpacademy-components/src/molecule/modal/index.js
```diff
@@ -76,12 +76,12 @@ Modal.propTypes = {
     left: PropTypes.shape({
       title: PropTypes.string,
       subtitle: PropTypes.string,
-      link: PropTypes.string
+      onClick: PropTypes.string
```
Fausto95:
```suggestion
      onClick: PropTypes.func
```

packages/@coorpacademy-components/src/molecule/modal/index.js
```diff
@@ -76,12 +76,12 @@ Modal.propTypes = {
...
     }),
     right: PropTypes.shape({
       title: PropTypes.string,
       subtitle: PropTypes.string,
-      link: PropTypes.string
+      onClick: PropTypes.string
```
Fausto95:
```suggestion
      onClick: PropTypes.func
```

packages/@coorpacademy-components/src/molecule/modal/index.js
```diff
@@ -13,9 +13,9 @@ import PropTypes from 'prop-types';
 import Provider from '../../atom/provider';
 import style from './style.css';
 
-const Choice = ({title, subtitle, link, children}) => {
+const Choice = ({title, subtitle, onClick, children}) => {
```
VincentCATILLON: why do we have two components in the same file ?

Fausto95:
> Because "history"

I feel you, sorry :(

PR #1676, opened 2020-01-23
Fix selectRules with multi answers for same slide

esa-coorp:

**Result and observation**

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [x] Unit testing

### DIFF
--- a/packages/@coorpacademy-progression-engine/src/rule-engine/select-rule.js
+++ b/packages/@coorpacademy-progression-engine/src/rule-engine/select-rule.js
@@ -1,5 +1,16 @@
 // @flow
-import {get, head, isEqual, filter, sortBy} from 'lodash/fp';
+import {
+  get,
+  head,
+  isEqual,
+  filter,
+  sortBy,
+  pipe,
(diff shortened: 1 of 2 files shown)


packages/@coorpacademy-progression-engine/src/rule-engine/select-rule.js
```diff
@@ -30,11 +41,13 @@ const matchWithState = (state: State) => (chapterRule: ChapterRule): boolean =>
       switch (target.scope) {
         case 'slide': {
           const {ref, field} = target;
-          const answerRecord = state.allAnswers.reverse().find(
-            (record): boolean => {
-              return record.slideRef === ref;
-            }
-          );
+
+          const answerRecord = pipe(
```
godu:
```js
          const answerRecord = pipe(
            reverse,
            uniqBy('slideRef'),
            find({slideRef: ref})
          )(state.allAnswers);
```

PR #1687, opened 2020-02-27
Use green as progression bar color instead of brands primary color

Fausto95:
https://trello.com/c/0Hnopfbd/1601-ux-end-of-primary-color-on-progression-card


packages/@coorpacademy-components/src/molecule/card-content/index.js
```diff
@@ -48,10 +48,10 @@ const ContentInfo = (
   context
 ) => {
   const {skin} = context;
-  const primaryColor = get('common.primary', skin);
   const whiteColor = get('common.white', skin);
+  const progressBarColor = '#3EC483';
```
silouone: mouai

PR #1690, opened 2020-03-02
Fix button components to support children

AdrieanKhisbe:
Component button mis en default... :smirk:


AdrieanKhisbe:
```
Successfully published:
 - @coorpacademy/components@9.9.7
 - @coorpacademy/player-web@3.5.9
```

PR #1697, opened 2020-03-24
Update video provider

Fausto95:
Update `VideoProvider` to also support `youtube` & `omniPlayer`


Fausto95:
i freaking missed this one

PR #1712, opened 2020-04-28
Fixing glitch on adaptive when switching slides or loading courses

alanlanglois:

This PR solves a little glitch that was visible when loading an adaptive chapter or switching between slides in adaptive questions. 
It also mutualises getProgressionsSteps, moving it from player-web to player-store.

**Result and observation**
Fix the glitch

- [ ] **Extra lib ?**

**Testing Strategy**

- [] Manual testing
- [x] Unit testing

### DIFF
--- a/packages/@coorpacademy-player-store/src/utils/state-extract.js
+++ b/packages/@coorpacademy-player-store/src/utils/state-extract.js
@@ -36,6 +36,7 @@ import type {
   ExitNode,
   Level,
   Lives,
+  ProgressionSteps,
   // eslint-disable-next-line no-unused-vars
   Resource,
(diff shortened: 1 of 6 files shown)


packages/@coorpacademy-player-store/src/utils/state-extract.js
```diff
@@ -502,6 +502,20 @@ export const getLives = (state: State): Lives => {
   };
 };
 
+export const getProgressionSteps = (state: State): {current: number, total: number} | null => {
```
Fausto95, line 505: Add Step type instead of typing the function directly

PR #1746, opened 2020-06-11
Add external content button

esa-coorp:

**Result and observation**

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [ ] Unit testing

### DIFF
--- a/packages/@coorpacademy-components/src/molecule/external-content-button/index.js
+++ b/packages/@coorpacademy-components/src/molecule/external-content-button/index.js
@@ -1,0 +1,53 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Provider from '../../atom/provider';
+import {EXTERNAL_CONTENT_ICONS} from '../../util/external-content';
+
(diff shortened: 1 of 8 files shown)


packages/@coorpacademy-components/src/molecule/external-content-button/index.js
```diff
@@ -0,0 +1,54 @@
...
+import PropTypes from 'prop-types';
+import Provider from '../../atom/provider';
+import {EXTERNAL_CONTENT_ICONS} from '../../util/external-content';
+
+import style from './style.css';
+
+class ExternalContentButton extends React.Component {
+  constructor(props, context) {
+    super(props, context);
+    this.handleClick = this.handleClick.bind(this);
```
Fausto95, line 11: No need to `bind` when using arrrow functions

esa-coorp: bien vu

PR #1748, opened 2020-06-12
Eslint

godu:


packages/@coorpacademy-components/src/atom/autocomplete/index.js
```diff
@@ -22,17 +21,22 @@ const Autocomplete = props => {
...
-  const handleSuggestionsSelected = (e, data) => onSuggestionSelected(data);
+  const handleChange = useMemo(() => e => onChange(e), [onChange]);
+  const handleBlur = useMemo(() => (e, selectedSuggestion) => onBlur(e, selectedSuggestion), [
+    onBlur
+  ]);
+  const handleSuggestionsFetchRequested = useMemo(() => e => onFetch(e), [onFetch]);
+  const handleSuggestionsClearRequested = useMemo(() => e => onClear(e), [onClear]);
+  const handleSuggestionsSelected = useMemo(() => (e, data) => onSuggestionSelected(data), [
+    onSuggestionSelected
+  ]);
```
Fausto95, line 39:
```suggestion
  const handleSuggestionsSelected = useMemo(() => (e, data) => onSuggestionSelected(data), [
    onSuggestionSelected,
    data
  ]);
```

Applicable everywhere, and please install https://www.npmjs.com/package/eslint-plugin-react-hooks so we get warned

packages/@coorpacademy-components/src/atom/cta/index.js
```diff
@@ -18,13 +35,13 @@ class CTA extends React.Component {
   }
 
   handleMouseEnter() {
-    return this.setState(prevState => ({
+    return this.setState(() => ({
```
Fausto95, line 38: Why even use an updater if we're not depending on previous state, just pass an object

packages/@coorpacademy-components/src/atom/autocomplete/index.js
godu: `data` is not in this scope. You misread the code.

Yeah, we could add `eslint-plugin-react-hooks`. You could do a PR on https://github.com/CoorpAcademy/eslint-plugin-coorpacademy for that.

packages/@coorpacademy-components/src/atom/cta/index.js
godu: I removed unused params. I could do more, but that you propose is not iso. There is a subtle difference between both. Mine is lazy, your's eager.
setState wait in the first place an updater function.
In this case, we don't care, the state isn't too complexe to pay attention of evaluation's performance.

packages/@coorpacademy-components/src/atom/autocomplete/index.js
Fausto95: Oh yea

packages/@coorpacademy-components/src/atom/input-html/index.js
```diff
@@ -26,31 +36,37 @@ class InputHtml extends React.Component {
...
-    };
-    const handlePreview = e => {
-      e.preventDefault();
-      this.setState(state => ({
-        text: state.text,
-        preview: !state.preview
-      }));
-    };
-    const iconContent = !this.state.preview ? (
+    const iconContent = !preview ? (
```
silouone, line 69: pk le ternaire est construit ds ce sens la :p ?

godu: Aucune idée. Mais c'était déjà le cas avant.

PR #1750, opened 2020-06-17
adding refresh button and title for table

jomaora:
**Result and observation**

Capture d’écran 2020-06-17 à 15 46 03

**Testing Strategy**

- Manual testing
- Unit testing

### DIFF
--- a/packages/@coorpacademy-components/src/organism/brand-table/index.js
+++ b/packages/@coorpacademy-components/src/organism/brand-table/index.js
@@ -7,7 +7,16 @@ import Loader from '../../atom/loader';
 import style from './style.css';
 
 const BrandTable = props => {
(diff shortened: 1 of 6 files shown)


packages/@coorpacademy-components/src/organism/brand-table/index.js
```diff
@@ -31,9 +40,17 @@ const BrandTable = props => {
       <Search {...search} />
     </div>
   ) : null;
+
+  const refreshView = onRefresh ? (
+    <div className={style.refresh} onClick={onRefresh}>
+      🔄 Refresh data
```
esa-coorp, line 46: props ou translate

jomaora: @esa-coorp done

PR #1761, opened 2020-07-21
[components] Fix animation & style ajustement for card list pagination

DjamelSoualmi:
Fix paging animation added in #1758 & updated style


packages/@coorpacademy-components/src/molecule/dashboard/cards-list/style.css
```diff
@@ -12,6 +12,7 @@
 .wrapper {
   padding: 16px 30px 8px;
   position: relative;
+  overflow: hidden;
```
AdrieanKhisbe, line 15: 👍

PR #1762, opened 2020-07-22
[components] Fix paging animation problem in CardList

DjamelSoualmi:
Fix paging animation problem
The feature was added in  #1758 
First fix #1761

**Testing Strategy**

- [ ] Manual testing
- [ ] Unit testing

### DIFF
--- a/packages/@coorpacademy-components/src/molecule/dashboard/cards-list/style.css
+++ b/packages/@coorpacademy-components/src/molecule/dashboard/cards-list/style.css
@@ -83,7 +83,7 @@
   font-family: 'Open Sans';
   font-size: 14px;
   font-weight: normal;
-  letter-spacing: 4px;
+  letter-spacing: 2px;
   color: dark;
   margin-right: 10px;
   line-height: 28px;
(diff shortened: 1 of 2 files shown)


packages/@coorpacademy-components/src/molecule/dashboard/cards-list/style.css
```diff
@@ -83,7 +83,7 @@
   font-family: 'Open Sans';
   font-size: 14px;
   font-weight: normal;
-  letter-spacing: 4px;
+  letter-spacing: 2px;
```
AdrieanKhisbe, line 86: 👍

packages/@coorpacademy-components/src/molecule/dashboard/cards-list/style.css
```diff
@@ -101,6 +101,7 @@
   margin-right: 0px;
   overflow: hidden;
   scroll-behavior: smooth;
+  transform: scale3d(1, 1, 1);
```
AdrieanKhisbe, line 104: 🧙‍♂️

PR #1771, opened 2020-08-04
external-content: Adding main content for loading/iframe

jomaora:

**Result and observation**

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [ ] Unit testing


packages/@coorpacademy-components/src/template/external-course/index.js
```diff
@@ -76,13 +89,19 @@ class ExternalCourse extends React.Component {
           </div>
           <Button
             type="button"
-            disabled={complete.disabled}
-            onClick={this.handleOnClick(complete)}
+            disabled={loading ? true : complete.disabled}
+            onClick={loading ? identity : this.handleOnClick(complete)}
```
Fausto95, line 93: If it's disabled when loading, no need for the noob function, not?

jomaora: one can never be too careful

PR #1892, opened 2021-06-07
fix: remove child from parent rather than removing the child itself cuz IE11 :)

Fausto95:

**Result and observation**

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [ ] Unit testing

### DIFF
--- a/packages/@coorpacademy-components/src/molecule/share/index.js
+++ b/packages/@coorpacademy-components/src/molecule/share/index.js
@@ -38,6 +38,7 @@ const copyTextViaExecCommand = text => {
     status = SHARE_STATUS.ERROR;
   }
 
+  // Element.remove is not supported on IE11 :/
   // eslint-disable-next-line unicorn/prefer-node-remove
   document.body.removeChild(textarea);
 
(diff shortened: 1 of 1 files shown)


DjamelSoualmi (approved):

packages/@coorpacademy-components/src/molecule/share/index.js
```diff
@@ -20,7 +20,9 @@ const copyTextViaExecCommand = text => {
 
   textarea.style.border = '0';
   textarea.style.padding = '0';
-  textarea.style.margin = '0';
+  textarea.style.margin = '0 auto';
```
esa-coorp, line 23: pas besoin surtout si c'est en display: none
```suggestion
```

packages/@coorpacademy-components/src/molecule/share/index.js
```diff
@@ -20,7 +20,9 @@ const copyTextViaExecCommand = text => {
...
+  textarea.style.display = 'none';
+  textarea.style.textAlign = 'center';
```
esa-coorp, line 25: idem
```suggestion
```

esa-coorp (changes requested):

esa-coorp (approved):

PR #1966, opened 2021-10-08
Refactor <Title /> atom component in order to be use on page or group form component

jomaora:

Use title component on form section: centralizing title and subtitle elements.

**Result and observation**

#### Title for page

Capture d’écran 2021-10-08 à 12 06 45

#### Title for form-group

Capture d’écran 2021-10-08 à 12 06 31

#### result in form-group organism

Capture d’écran 2021-10-08 à 12 08 04

**Testing Strategy**

- Already covered by tests
- Manual testing
- Unit testing


silouone (approved):
pas de default donc design pété (mm ps de style appliqué il me semble) si pas de type passé pour une raison.

PR #2042, opened 2021-11-25
Fix css for buttons, removes bullet point menu interference

StefanoA1:
Using the class .button in css file impact other components, storyBook puts all the css in only one file so they become impacted by the pointer events modif.

Adds further explanations for the css.

**Result and observation**

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [ ] Unit testing

### DIFF
--- a/packages/@coorpacademy-components/src/molecule/bullet-point-menu-button/style.css
+++ b/packages/@coorpacademy-components/src/molecule/bullet-point-menu-button/style.css
@@ -10,7 +10,7 @@
   position: relative;
 }
 
-.bulletPointWrapper [data-name="bullet-point-button"] {
+.bulletPointWrapper > button {
   position: relative;
   cursor: pointer;
  }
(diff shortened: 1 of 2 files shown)


packages/@coorpacademy-components/src/molecule/bullet-point-menu-button/style.css
```diff
@@ -10,7 +10,7 @@
   position: relative;
 }
 
-.bulletPointButton {
+.bulletPointWrapper [data-name="bullet-point-button"] {
```
StefanoA1, line 13: je ne peux pas wrapper le button pour faire un class apart car j'ai besoin du combo siblings + focus sur le button, donc je target le bouton avec le data-name

DjamelSoualmi (approved):

jomaora (approved):

PR #2056, opened 2021-12-01
Fix select tooltip

esa-coorp:
https://user-images.githubusercontent.com/15608581/144214303-7658d084-0c2b-445b-945a-13b41812e849.mp4

**Result and observation**

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [ ] Unit testing

### DIFF
--- a/packages/@coorpacademy-components/src/molecule/select-multiple/style.css
+++ b/packages/@coorpacademy-components/src/molecule/select-multiple/style.css
@@ -235,6 +235,7 @@
   height: 44px;
   padding: 11px 85px 11px 16px;
   background-color: cm_grey_100;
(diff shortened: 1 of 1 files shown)


packages/@coorpacademy-components/src/molecule/select-multiple/style.css
```diff
@@ -235,6 +235,7 @@
   height: 44px;
   padding: 11px 85px 11px 16px;
   background-color: cm_grey_100;
+  overflow: visible;
```
Fausto95, line 238: 🤯😱 thanks

PR #2058, opened 2021-12-08
Fix Select Multiple Choice(weird bug, weird fix :/)

Fausto95:
Fix a weird bug that was happening whenever we were using a multiple choice

### DIFF
--- a/packages/@coorpacademy-components/src/molecule/select-multiple/index.js
+++ b/packages/@coorpacademy-components/src/molecule/select-multiple/index.js
@@ -72,7 +72,7 @@ const SelectMultiple = (
     placeholder,
     description,
     hint,
-    multiple = true,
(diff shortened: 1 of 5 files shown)


godu (approved):

StefanoA1:
ah weird bug indeed, c'est bon 👍

StefanoA1 (approved):

packages/@coorpacademy-components/src/molecule/select-multiple/index.js
```diff
@@ -72,7 +72,7 @@ const SelectMultiple = (
     placeholder,
     description,
     hint,
-    multiple,
+    multiple = true,
```
Fausto95, line 75:
```suggestion
    multiple,
```

PR #2169, opened 2022-03-22
fix right icon - to center

StefanoA1:
 - Fixes selected option color on Dropdown

**Result and observation**
before:
Screenshot 2022-03-14 at 17 03 04

after:
Screenshot 2022-03-14 at 17 03 11

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [ ] Unit testing

### DIFF
--- a/packages/@coorpacademy-components/src/atom/review-header-step-item/style.css
+++ b/packages/@coorpacademy-components/src/atom/review-header-step-item/style.css
@@ -45,14 +45,14 @@
   color: white;
   width: 12px;
   height: 10px;
-  margin: 2px 0px 0px 1px;
+  margin: 4px 0px 0px 1px;
   transition: all 0.25s ease;
 }
 
 .currentRightIcon {
   width: 16px;
   height: 14px;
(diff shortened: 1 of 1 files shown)


packages/@coorpacademy-components/src/atom/review-header-step-item/style.css
```diff
@@ -45,14 +45,14 @@
   color: white;
   width: 12px;
   height: 10px;
-  margin: 2px 0px 0px 1px;
+  margin: 4px 0px 0px 1px;
```
adamska27, line 48: moi je supprimerai les margin et je mettrai du flex sur le container au dessus

adamska27: sur 
````
.value {
   display: flex;
   justify-content: center;
   align-items: center;
}
`````

adamska27 (approved):

PR #2237, opened 2022-06-15
Review add init step for review

jomaora:
https://trello.com/c/TCjT1J7z/2642-orprogression-engine-add-initstepforreview-function

This PR adds the `computeInitialStepForReview` function that allow `createProgression` to build the initialAction move for a progression of type review.

Also, the name used for the type of `AvailableContent`, that is an Array of `ChapterContent` was changed for a more generic one `ContentContainer`

```js
{
  ref: string,
  slides: Array<Slide>,
  rules: Array<ChapterRule> | null
}
```

This type is gonna be used also for review mode. `ref` would be the skill ref and slides is gonna be an array of one slide.

**Result and observation**

A new progression of type review is created with its initial step.
The compute of the `actions` property on the `createProgression` function was refactored to handle the review engine.

**Testing Strategy**

- Already covered by tests
- Unit testing

### DIFF
--- a/packages/@coorpacademy-progression-engine/src/compute-next-step/compute-next-step.js
+++ b/packages/@coorpacademy-progression-engine/src/compute-next-step/compute-next-step.js
@@ -23,7 +23,7 @@ import type {
   Answer,
   AnswerRecord,
   AvailableContent,
-  ChapterContent,
+  ContentContainer,
   Config,
   Content,
   Slide,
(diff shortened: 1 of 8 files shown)


StefanoA1 (approved):

packages/@coorpacademy-progression-engine/src/compute-next-step/compute-next-step.js
```diff
@@ -357,4 +357,22 @@ const computeNextStep = (
...
+    return null;
+  }
+
+  return {
+    nextContent: {
+      type: 'slide',
+      ref: nextSlide._id
+    },
+    instructions: null,
+    isCorrect: false
```
esa-coorp, line 376: pourquoi false?

jomaora: @esa-coorp c'est un code itératif, la vrai valeur sera pris du state mais pour la création le state est null. La logique pour le isCorrect viendra dans le ticket de computeNextContentAfterAnswer

esa-coorp (approved):

godu (approved):

PR #2268, opened 2022-07-19
Migrating touchable

chrisdugne:

moving `Touchable` here from https://github.com/CoorpAcademy/mobile


chrisdugne:
moved to https://github.com/CoorpAcademy/components/compare/mobile-touchable?expand=1

PR #2342, opened 2022-08-30
[app review] set skin via new WebContext

jomaora:
https://trello.com/c/W34NVprZ/2703-app-review-set-skin-for-player-review

On components, skin is get from context via the legacy format (second parameter on each Component fonction, and defined via the `<Provider />` atom). This is the way that the `player-web` set the colors and the translate fonctions.

Before this PR, the `app-review` was unable to set skin brand. As the `app-review` is using the hooks from React, a specific context: [WebContext](https://github.com/CoorpAcademy/components/pull/2342/files#diff-13d23068b0243b8203bfa291cfe441867074489f6c3aeaf16a8b9904bef360c0) was added to allow this. 

This context will work aside the legacy context to guarantee the `player-web` behavior. Modifications were applied to the different type of questions, where access to primaryColor is necessary.

**Result and observation**

The PR fixes the error seen on the Components debugger tool. Skin being not set, we have this on the console when we tried to inspect the ReviewSlide component (this was rename on the current PR).

**before**
Capture d’écran 2022-09-02 à 11 36 39
**after**
Capture d’écran 2022-09-02 à 11 40 49

### qcm
**before**
Capture d’écran 2022-09-02 à 11 20 27

**after**
qcm

### qcmGraphic
**before**
Capture d’écran 2022-09-02 à 11 26 26

**after**
qcm-Graphic

### qcmDrag
**before**
Capture d’écran 2022-09-02 à 11 29 46

**after**
qcmDrag

### slider
**before**
Capture d’écran 2022-09-02 à 11 27 35

**after**
slider

**Breaking changes ?**  
We've decided with @esa-coorp and @audric-coorp to stop using the primaryColor on the background of the `<Range/>`component. This is still code using the legacy class format for a component and it's migration was taking a lot of time. This component should be migrated when possible.

**Testing Strategy**

- Already covered by tests
- Manual testing
- Unit testing


chrisdugne (approved):
pour moi il serait plus clair de remonter les contextes dans un `src/contexts` plutot que dans `src/atom/provider`

StefanoA1 (approved):

esa-coorp (approved):

PR #2360, opened 2022-09-07
[app-review] Fix height on review player

jomaora:
https://trello.com/c/8gmr8Ob6/2751-components-fixer-laffichage-du-player-prendre-la-hauteur-de-l%C3%A9cran

App-review components where not taking the height of the screen to be display. It makes that the player is shown with vertical scroll when the screen is little.

This PR fixes the css to adjust the player to the screen.

**Result and observation**

### Before

height-review-player-before

### After
height-review-player-after

### On IE 11

Capture d’écran 2022-09-13 à 15 19 08

**Testing Strategy**

- Manual testing


packages/@coorpacademy-components/src/organism/review-stacked-slides/style.css
```diff
@@ -153,22 +150,15 @@
...
-    height: 670px;
     max-width: 335px;
   }
-
   .slideBase {
-    height: 654px;
+    height: calc(100vh - 185px);
   }
-}
+}
```
StefanoA1, line 164: retour à la ligne

EmelineLeduc (approved):

PR #2373, opened 2022-09-13
updated app-review sandbox

chrisdugne:

should allow sandbox to run for ie11 (current build format is `cjs` = for  node environment)
https://esbuild.github.io/api/#format-iife


jomaora (changes requested):
Marche pas sur IE

chrisdugne:
esbuild was removed to use webpack #2380

PR #2462, opened 2022-10-25
Add skill to review type

antonia2929:
Part of this trello card : https://trello.com/c/swIZN13h/2800-app-review-afficher-le-nom-de-la-skill-choisie-dans-le-player

The goal is to get the skill name in player review slide.
Skills in dashboard used in skills views are returned by api-review.
For this card, the required skill will be requested to an existed endpoint in mooc.
So new type has been created with the required properties skillRef and name.

Add new types to app-review and refacto Skill type

- Skill change to SkillsToReview
- New Skill type only have skillRef and name. This type will be used to get the name in player review slide.

**Result and observation**
New type Skill and refacto SkillsToReview

**Testing Strategy**
- Already covered by tests

### DIFF
--- a/packages/@coorpacademy-app-review/src/types/common.ts
+++ b/packages/@coorpacademy-app-review/src/types/common.ts
@@ -154,6 +154,11 @@ export type SkillsToReview = {
   name: string;
 };
 
+export type Skill = {
+  name: string;
+  skillRef: string;
+};
+
 export type Services = {
   fetchSkills(token: string): Promise<SkillsToReview[]>;
   fetchSlide(slideRef: string, token: string): Promise<SlideFromAPI | void>;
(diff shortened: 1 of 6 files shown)


packages/@coorpacademy-app-review/src/types/common.ts
```diff
@@ -147,15 +147,20 @@ export type CorrectionFromAPI = {
   corrections: CorrectedChoice[];
 };
 
-export type Skill = {
+export type SkillsToReview = {
```
StefanoA1, line 150: est-ce que tu peux mettre ça en singulier stp? skill**s** donne une idée de collection plutôt
```suggestion
export type SkillToReview = {
```

antonia2929: @StefanoA1 ça marche

StefanoA1 (approved):

PR #2476, opened 2022-11-07
update fetchSkill service

antonia2929:
Part of this issue : https://trello.com/c/swIZN13h/2800-app-review-afficher-le-nom-de-la-skill-choisie-dans-le-player

- fix the fetchSkill service to return the first element of the returned array of skills.

**Result and observation**

**Testing Strategy**
- [x] Unit testing

### DIFF
--- a/packages/@coorpacademy-app-review/src/services/fetch-skill.ts
+++ b/packages/@coorpacademy-app-review/src/services/fetch-skill.ts
@@ -3,7 +3,7 @@ import decode from 'jwt-decode';
 import {JWT, Skill} from '../types/common';
 import {toJSON} from './tools/fetch-responses';
 
-export const fetchSkill = async (skillRef: string, token: string): Promise<Skill | void> => {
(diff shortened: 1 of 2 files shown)


packages/@coorpacademy-app-review/src/services/fetch-skill.ts
```diff
@@ -8,5 +8,6 @@ export const fetchSkill = async (skillRef: string, token: string): Promise<Skill
   const response = await crossFetch(`${host}/api/v2/skills?conditions={"ref":"${skillRef}"}`, {
     headers: {authorization: token}
   });
-  return toJSON<Skill>(response);
+  const skills = await toJSON<Skill[]>(response);
+  return skills[0];
```
scalabw, line 12: What would happen if there is an error with the API call ?

don't forget to check that skills[0] exist 😊

jomaora (approved):

StefanoA1 (approved):

PR #2482, opened 2022-11-08
[review] Fix compute next step for review

jomaora:
https://trello.com/c/xZaKvARK/2828-or-bug-double-slide-impossibilit%C3%A9-de-quitter-le-review-mode

When a slide is answered on the review mode, its progression it's sent via the kinesis progression stream to be handle by the `recallUpdate` lambda. This treatment should mark that the slide has a new recallDate and avoid to propose it as an available question to be answered on the review mode (it's has been already answered)

But, `recallUpdate` can lag, and make that `getSlide` lambda would propose again the already answered slide.
To avoid this problem, the group of available slides would be filter to exclude the already answered slides (on `state.slides`).

**Testing Strategy**

- Unit testing

### DIFF
--- a/packages/@coorpacademy-progression-engine/src/compute-next-step/compute-next-step.js
+++ b/packages/@coorpacademy-progression-engine/src/compute-next-step/compute-next-step.js
@@ -388,7 +388,7 @@ const getNextPendingSlide = (
 export const computeNextStepForReview = (
   config: Config,
   _state: State | null,
-  availableContent: AvailableContent,
+  _availableContent: AvailableContent,
   partialAction: PartialAction
 ): Result => {
   const action = extendPartialAction(partialAction, _state);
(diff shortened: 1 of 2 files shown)


antonia2929 (approved):

packages/@coorpacademy-progression-engine/src/compute-next-step/compute-next-step.js
```diff
@@ -364,9 +365,17 @@ const getNextSlide = (
 ): Slide | null => {
   if (!state) return get(['0', 'slides', '0'], availableContent);
 
+  // We filter slides in orther to exclude already answered slides.
```
scalabw, line 368: typo ?
"orther" "order"

scalabw (approved):

esa-coorp (approved):

StefanoA1 (approved):

PR #2552, opened 2022-12-29
Add dropdown icon arialabel +trad

StefanoA1:

- Adds dropdown icon arialabel +trad, adds tests
- Adds translate context for mobile components + removes isRequired from context translate as for shared components (native) the context does not need to provide a context forcefully.

**Testing Strategy**

- [x] Unit testing

### DIFF
--- a/packages/@coorpacademy-components/locales/en/global.json
+++ b/packages/@coorpacademy-components/locales/en/global.json
@@ -20,6 +20,7 @@
   "Create a new cohort": "Create a new cohort",
   "Delete": "Delete",
   "Draft": "Draft",
+  "drop_down_icon": "Dropdown icon, to open a list of available answers",
   "Edit": "Edit",
   "external_content_article": "DOCUMENTS",
   "external_content_scorm": "ANIMATED RESOURCES",
--- a/packages/@coorpacademy-components/locales/en/global.json
+++ b/packages/@coorpacademy-components/locales/en/global.json
@@ -20,7 +20,7 @@
   "Create a new cohort": "Create a new cohort",
   "Delete": "Delete",
   "Draft": "Draft",
-  "drop_down_icon": "Dropdown icon, to open a list of available answers",
+  "drop_down_icon": "Dropdown to open the list of available answers",
   "Edit": "Edit",
   "external_content_article": "DOCUMENTS",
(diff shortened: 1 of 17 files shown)


StefanoA1:
[[ReviewPlayer & ClassicPlayer] Dropdown icon is missing an aria-label (Slides)](https://trello.com/c/GTHjLQ2S/114-reviewplayer-classicplayer-dropdown-icon-is-missing-an-aria-label-slides)

packages/@coorpacademy-components/src/atom/select/index.js
```diff
@@ -42,7 +42,7 @@ const Select = (props, legacyContext) => {
     modified = false,
     error = false,
     title: propTitle,
-    moreAriaLabel
+    'aria-label': ariaLabel
```
StefanoA1, line 45: changed prop's name because the icon can have many trads now (more, dropdown answer, possibly others later).

packages/@coorpacademy-components/locales/en/global.json
```diff
@@ -20,6 +20,7 @@
   "Create a new cohort": "Create a new cohort",
   "Delete": "Delete",
   "Draft": "Draft",
+  "drop_down_icon": "Dropdown icon, to open a list of available answers",
```
DjamelSoualmi, line 23:
```suggestion
  "drop_down_icon": "Dropdown to open the list of available answers",
```
le user sais déjà que c'est un svg

DjamelSoualmi: @audric-coorp

DjamelSoualmi (changes requested):

esa-coorp (approved):

packages/@coorpacademy-components/src/atom/provider/index.js
```diff
@@ -68,7 +68,7 @@ class Provider extends React.Component {
   static childContextTypes = {
     history: historyShape,
     skin: skinShape,
-    translate: translateShape.isRequired,
+    translate: translateShape,
```
StefanoA1, line 71: I removed it because it is not forcefully required anymore as in mobile's case, it is not compulsory present on native components that used shared basic components

antonia2929 (approved):

PR #2604, opened 2023-01-24
Review change actions end progression

jomaora:
https://github.com/CoorpAcademy/components/pull/2604#issue-1555289040

In order to avoid the case where the user can create a progression and no more skills are avaliable, we move the fetch skills and rank to the next "Continue" on the last correction popin, before the congratulations view.

**Result and observation**

**Before**

before-actions-on-last-answer

**After**
Same behavior as before for the users. But actions are now dispatched when click on the correction popin

actions-during-next-slide

**Testing Strategy**

- Manual testing
- Unit testing

### DIFF
--- a/packages/@coorpacademy-app-review/src/actions/test/create-test-store.ts
+++ b/packages/@coorpacademy-app-review/src/actions/test/create-test-store.ts
@@ -1,5 +1,6 @@
 import type {ExecutionContext} from 'ava';
 import constant from 'lodash/fp/constant';
+import isEqual from 'lodash/fp/isEqual';
 import {AnyAction, applyMiddleware, compose, createStore, Dispatch, Middleware, Store} from 'redux';
 import thunk from 'redux-thunk';
 import type {ThunkOptions} from '../../types/common';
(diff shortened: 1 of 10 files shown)


packages/@coorpacademy-app-review/src/actions/test/create-test-store.ts
```diff
@@ -10,6 +11,11 @@ const assertActionsMiddleware = (t: ExecutionContext, ACTIONS: AnyAction[]): Mid
     const expectedAction = ACTIONS.shift();
     if (!expectedAction) throw new Error(`Missing {type: "${action.type}"} action`);
     t.deepEqual(expectedAction, action);
+    if (!isEqual(expectedAction, action)) {
+      throw new Error(
+        `Difference between received action {type: "${action.type}"}, an expected action ${expectedAction.type}. Check payloads.`
```
scalabw, line 16: So cool ! 👍

scalabw (approved):

EmelineLeduc (approved):

PR #2631, opened 2023-02-13
Adapt header review step item to support bulk import design

youssefezzahi96:
[Ticket
](https://trello.com/c/YbO0HtBE/1188-composant-inspect)**Purpose 1:** 
in order to respect bulk import design (inspect page), we need to have these graphic objects: 
image image image
header review step item uses similar objects, but with different styles and different aria-labels, so the purpose is to make the component (header review step item) usable for both cases (bulk import and header review steps). 

**To do this:** 
- I added other icon props (valid, invalid, errors-number)
- Depending on these new icon props, i specified style and aria-label 
- I changed header review step item name to status step to be more generic name.
- I added bulk import status items fixtures

**Purpose 2:** 
To correct existing header review step item Accessibility (RGAA):
existing problems: 
- aria-describedby should have an id of an html element as value not a description
- having aria-label and aria-describedby on the wrapping tag (span) with specified locales makes information duplication, for example we would have  (aria-label: question 1 , aria-describedby: question 1 is not answered), so question 1 is a duplicated information
- aria-label should be on svg not on the wrapping tag (span in our case)
- if there is no svg (no-answer case), in this case, the wrapping tag could have aria-label, bu needs also img="role"

**To correct this,** 
depending on cases: 
-with svg (right, wrong, valid, invalid): i specify on the svg aria-label (ex: current question 1 is correct) and role="img"
-without svg (no-answer, errors-number) i specify arround the svg (span) the aria-label (ex: current question 1 is not answered) and role="img". => much simpler and all informations are in aria-label.
-I updated trads


packages/@coorpacademy-components/src/atom/status-item/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+  let child;
+  let contentAriaLabel;
+  switch (icon) {
+    case 'no-answer':
+      child = value;
+      contentAriaLabel = translate('review_header_step_item.not_answered_question', {
+        current: current ? 'current ' : '',
+        headerStepValue: value
+      });
+      break;
```
DjamelSoualmi, line 23: j'ai pas la traduction sur storybook !

packages/@coorpacademy-components/src/atom/status-item/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+      );
+      break;
+
+    default:
+      return null;
+  }
+
+  return (
+    <span
+      {...(icon === 'no-answer' || icon === 'errors-number'
```
DjamelSoualmi, line 84: dans le cas de answer correct et incorrect  le span dois avoir   un role status ou alerte

DjamelSoualmi (changes requested):

DjamelSoualmi (approved):

PR #2633, opened 2023-02-13
adapt button link icon only to support bulk import design

youssefezzahi96:
[Ticket](https://trello.com/c/YbO0HtBE/1188-composant-inspect)
Purpose 1:
in order to respect bulk import design (inspect page), we need to have these graphic objects:
image image image

"button link icon only" uses similar objects, but with different styles, so the purpose is to make the component (button link icon only ) usable for bulk import.

**To do this:**

- I added "down" icon to the possible set icon props
- I created needed classes "bulkArrowDown" "bulkArrowUp" 
- These classes could be set in props (example : icon: 'down', className: bulkArrowDown)
- I created fixtures


silouone (approved):
lgtm, on verifiera le 'no impact' sur review

GhaziCoorpacademy (approved):

PR #2636, opened 2023-02-14
[app-review] improved UX on review slide validation

chrisdugne:

After having validated, the user could still trigger options in the slide.
This update alloes to act immediately once the validate button is triggered:

- removing textInput focus,
- adding a hidden BG to capture touches,
- triggering the real 'validate' action on the next frame, to be sure the locking BG appears the quickest possible.

before:
before

after:
after

### DIFF
--- a/packages/@coorpacademy-components/src/organism/review-slide/test/on-press.test.tsx
+++ b/packages/@coorpacademy-components/src/organism/review-slide/test/on-press.test.tsx
@@ -5,7 +5,7 @@ import Slide from '../index.native';
 import mockedGestureEvent from '../../../test/helpers/mocked-gesture-event';
 import fixtures from './fixtures/initial-state';
 
-test('should handle onPress on validate button', t => {
+test('should handle onPress on validate button', async t => {
   t.plan(1);
 
   const props = fixtures.props;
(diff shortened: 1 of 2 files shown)


chrisdugne:
[[OR] App mobile - Clic possible sur les réponses après validation de la question](https://trello.com/c/vqgEuE7k/2977-or-app-mobile-clic-possible-sur-les-r%C3%A9ponses-apr%C3%A8s-validation-de-la-question)

jomaora (approved):

packages/@coorpacademy-components/src/organism/review-slide/test/on-press.test.tsx
```diff
@@ -18,4 +18,9 @@ test('should handle onPress on validate button', t => {
   const buttonValidate = getByTestId('slide-validate-button-0');
 
   fireEvent.press(buttonValidate, mockedGestureEvent);
+
+  // wait for onclick to be triggered
+  await new Promise(resolve => {
+    setTimeout(resolve, 50);
```
scalabw, line 24: on the code of the component the time is 20ms but here it is 50ms is there any reason for that ?

scalabw, line 24: on the code of the component the time is 20ms but here it is 50ms is there any reason for that ?

chrisdugne: yes, not to have the test failing, rather to wait few more ms to be sure the function was called

scalabw (approved):

PR #2643, opened 2023-02-16
[RGAA] update svg generator in order to take other roles if set

youssefezzahi96:
purpose of the PR: 
the problem is that all generated svg would have img role by default in case we set aria-label or alt, but in some cases, we would prefer to have an other role (status as example).
so this PR offers this possibility.

### DIFF
--- a/packages/@coorpacademy-nova-icons/scripts/generate-icons.js
+++ b/packages/@coorpacademy-nova-icons/scripts/generate-icons.js
@@ -122,7 +122,7 @@ const template =
   `
       : templateAlias.ast`
     ${importsAlias}
-    const ${componentName} = _props => {const {'aria-label':ariaLabel, role, alt, ...rest} = _props; const props = {...rest, ...((ariaLabel || alt) ? {role: role ?? 'img', 'aria-label':ariaLabel, alt} : {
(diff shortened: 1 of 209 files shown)


packages/@coorpacademy-nova-icons/scripts/generate-icons.js
```diff
@@ -122,7 +122,7 @@ const template =
   `
       : templateAlias.ast`
     ${importsAlias}
-    const ${componentName} = _props => {const {'aria-label':ariaLabel, alt, ...rest} = _props; const props = {...rest, ...((ariaLabel || alt) ? {role: 'img', 'aria-label':ariaLabel, alt} : {
+    const ${componentName} = _props => {const {'aria-label':ariaLabel, role, alt, ...rest} = _props; const props = {...rest, ...((ariaLabel || alt) ? {role: role ?? 'img', 'aria-label':ariaLabel, alt} : {
```
StefanoA1, line 125:
```suggestion
const ${componentName} = _props => {const {'aria-label':ariaLabel, alt, ...rest} = _props; const props = {...rest, ...((ariaLabel || alt) ? {role: role || 'img', 'aria-label':ariaLabel, alt} : {
```

StefanoA1 (changes requested):

DjamelSoualmi (approved):

PR #2666, opened 2023-02-28
fixing atom.range.handle

chrisdugne:

issue:
- `useMemo` is triggered before the first render, meaning `ref` is not created

```
Uncaught TypeError: Cannot read properties of undefined (reading 'addEventListener')
    at eval (hammer.js:237:14)
```

- solution: calling `useMemo` once again when the render is done and the `ref` is updated


godu (approved):

scalabw (approved):
It Works !

PR #2673, opened 2023-03-06
Fix hoc animation

scalabw:
https://trello.com/c/TQeuy4Nb/3004-mooccours-pop-en-boucle-%C3%A0-l%C3%A9cran-de-fin

Before:
Kapture 2023-03-06 at 14 16 02
After:
Kapture 2023-03-06 at 14 19 27

This PR fix the animation of the stars and rank at the end of a course, It was supposed to bounce only once but that wasnt the case.

### DIFF
--- a/packages/@coorpacademy-components/src/hoc/transition/index.tsx
+++ b/packages/@coorpacademy-components/src/hoc/transition/index.tsx
@@ -9,14 +9,14 @@ const Child = ({child, onAnimationEnd, className, name}: ChildProps) => {
     props: {className: propClassName}
   } = child;
 
-  const handlerAnimationEnd = useCallback(() => {
+  const handleAnimationEnd = useCallback(() => {
     return onAnimationEnd(name);
   }, [onAnimationEnd, name]);
 
   return React.cloneElement(child, {
     className: classnames(propClassName, className),
-    onTransitionEnd: handlerAnimationEnd,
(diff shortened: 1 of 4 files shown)


StefanoA1 (changes requested):

StefanoA1:
+, if you could add coverage for `handleChange` lines (nice to have, not required)

packages/@coorpacademy-components/src/hoc/transition/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  const handlerAnimationEnd = useCallback(() => {
+    return onAnimationEnd(name);
+  }, [onAnimationEnd, name]);
+
+  return React.cloneElement(child, {
+    className: classnames(propClassName, className),
+    onTransitionEnd: handlerAnimationEnd,
+    onAnimationEnd: handlerAnimationEnd
+  });
```
StefanoA1, line 20:
```suggestion
  const handleAnimationEnd = useCallback(() => {
    return onAnimationEnd(name);
  }, [onAnimationEnd, name]);

  return React.cloneElement(child, {
    className: classnames(propClassName, className),
    onTransitionEnd: handleAnimationEnd,
    onAnimationEnd: handleAnimationEnd
  });
```

StefanoA1 (approved):

PR #2699, opened 2023-03-22
Fixes input that should use display: none; instead of opacity

StefanoA1:

Fixes input that should use display: none; instead of opacity

**Result and observation**
Before:
Screenshot 2023-03-22 at 16 56 24

After:
Screenshot 2023-03-22 at 16 56 16

**Testing Strategy**

- [x] Manual testing
- [ ] Unit testing


StefanoA1:
[Fix: use display: none instead of opacity: 0, when hiding a component - E2E](https://trello.com/c/1oVrEYmm/164-fix-use-display-none-instead-of-opacity-0-when-hiding-a-component-e2e)

PR #2737, opened 2023-05-22
added mode=scorm to microlearning

GhaziCoorpacademy:
https://trello.com/c/xtJFnLdh/1272-return-button-player-coorp-content-scorm-should-not-be-clickable
- When scorm mode: Removed back arrow + hover + cursor
Before:
before
After:
after

**Result and observation**

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [ ] Unit testing


GhaziCoorpacademy:
[return button player coorp content scorm should not be clickable](https://trello.com/c/xtJFnLdh/1272-return-button-player-coorp-content-scorm-should-not-be-clickable)

silouone (approved):

PR #2741, opened 2023-05-25
Updating browserlist setup

chrisdugne:

we have local issues running `npx ava` on package `@coorpacademy/component`, requiring to update the browserslist  config.

- added latest `browserslist` and `caniuse-lite` within devdeps


chrisdugne:
does not work 😿

PR #2773, opened 2023-07-28
Popin with cards

antonia2929:
Trello card : [[component] New component Popin with cards (desktop + mobile)](https://trello.com/c/t9IGxywY/3131-component-new-component-popin-with-cards-desktop-mobile)

- Adapt the `CmPopin` property to take a list of items like CardsGrid, and update the header to have a title property
- Add `PopinWithCard` fixtures

**Result and observation**
Kapture 2023-08-08 at 18 24 57

**Testing Strategy**

- [ ] Manual testing
- [ ] Unit testing

### DIFF
--- a/packages/@coorpacademy-components/src/molecule/cm-popin/index.js
+++ b/packages/@coorpacademy-components/src/molecule/cm-popin/index.js
@@ -60,7 +60,7 @@ const CMPopin = props => {
           </div>
         );
       }
-      return <img className={style.headerBackground} src={header} />;
+      return <img className={style.headerBackground} src={header.backgroundImage} />;
     }
     if (mode === 'cookie')
       return (
--- a/packages/@coorpacademy-components/src/molecule/cm-popin/index.js
+++ b/packages/@coorpacademy-components/src/molecule/cm-popin/index.js
@@ -33,7 +33,6 @@ const CMPopin = props => {
     listBtnSwicth,
     items
   } = props;
-
   const logo = {
     AlertDiamond,
     WindowUpload,
@@ -118,6 +117,10 @@ const CMPopin = props => {
         </div>
       );
     })(listBtnSwicth);
+  };
+  const renderItems = () => {
+    const {type, list} = items;
+    return type === 'content' ? <CardsGrid {...list} /> : null;
   };
   return (
     <div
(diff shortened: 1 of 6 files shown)


packages/@coorpacademy-components/src/molecule/cm-popin/index.js
```diff
@@ -190,6 +217,7 @@ const CMPopin = props => {
             ) : null}
           </div>
         ) : null}
+        {items ? <div className={style.itemsContainer}>{renderItems()}</div> : null}
```
jomaora, line 220: I think that items should be shown before the buttons (even, if it's not what we're gonna use, if someone someday will activate the buttons, we'll have this

Capture d’écran 2023-08-10 à 11 46 14

jomaora (changes requested):

jomaora (approved):

packages/@coorpacademy-components/src/molecule/cm-popin/types.ts
```diff
@@ -36,7 +42,17 @@ const propTypes = {
...
-  listBtnSwicth: PropTypes.arrayOf(PropTypes.shape(InputSwitch.propTypes))
+  listBtnSwicth: PropTypes.arrayOf(PropTypes.shape(InputSwitch.propTypes)),
+  items: PropTypes.shape({
+    type: PropTypes.string,
+    list: PropTypes.shape(CardsGrid.propTypes)
+  })
+};
+
+type PopinHeaderProps = {
+  title: typeof Title.propTypes;
```
StefanoA1, line 53: `propTypes`  shouldn't be used as types (if you check the inner code, the return type is not the same base TypeScript Type), ex:
Screenshot 2023-08-10 at 15 01 24
TitleProps needs to be defined with valid types
```suggestion
  title: TitleProps;
```

StefanoA1 (changes requested):

packages/@coorpacademy-components/src/molecule/cm-popin/index.js
```diff
@@ -103,6 +124,10 @@ const CMPopin = props => {
       );
     })(listBtnSwicth);
   };
+  const renderItems = () => {
+    const {type, list} = items;
+    return type === 'content' ? <CardsGrid {...list} /> : null;
+  };
```
StefanoA1, line 130:
```suggestion
const {type, list} = items;
  const renderItems = useCallback(() => {
    return type === 'content' ? <CardsGrid {...list} /> : null;
  }, [type, list]);
```

PR #2820, opened 2023-11-24
Faurecia-vr video provider support

esa-coorp:
A I-Frame will be shared to Coorp Academy. This I-frame give the possibility to be played in VR and 
in fullscreen. So for that this i-frame will contain this following option to be run on navigator: 
 
allow="xr-spatial-tracking;fullscreen"

Capture d’écran 2023-11-24 à 13 30 09

**Result and observation**

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [ ] Unit testing

### DIFF
--- a/packages/@coorpacademy-components/src/molecule/video-iframe/index.js
+++ b/packages/@coorpacademy-components/src/molecule/video-iframe/index.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import qs from 'qs';
-import {noop} from 'lodash/fp';
+import {noop, includes} from 'lodash/fp';
 import Provider from '../../atom/provider';
 import {SrcPropType} from '../../util/proptypes';
 import style from './style.css';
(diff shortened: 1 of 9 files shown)


EmelineLeduc (approved):

packages/@coorpacademy-components/src/molecule/video-iframe/index.js
```diff
@@ -47,18 +55,15 @@ class VideoIframe extends React.Component {
   render() {
     const {type, id, url, autoplay = false, width = '100%', height = '400px'} = this.props;
     const src = getUrl({url, type, id, autoplay});
-
+    const properties = {
+      allow: includes(['faurecia-vr'], type) ? 'xr-spatial-tracking;fullscreen' : null,
```
adamska27, line 59: c'est plutôt l'inverse, comme tu as fait plus bas?
`includes(value, collection)`
image
 Quand je teste il supporte pas un array en 1er param
```suggestion
      allow: includes(type, ['faurecia-vr']) ? 'xr-spatial-tracking;fullscreen' : null,
```

adamska27 (approved):

PR #2846, opened 2024-04-19
Add Learner Skill Card component

HuyIT9898:
l**Result and observation**
Learner skill card with Review and Explore Buttons

image

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [ ] Unit testing


packages/@coorpacademy-components/src/atom/icon/index.js
```diff
@@ -63,7 +63,7 @@ const Icon = React.memo(function Icon({
       <FontAwesomeIcon
         icon={`fa-${iconName}`}
         color={effectiveIconColor}
-        font-size={effectiveSize.faSize}
+        fontSize={effectiveSize.faSize}
```
AdrieanKhisbe, line 66: (style comment: I personaly prefer kebab-case for attribute names in jsx, but that's a detail, and discutable)

packages/@coorpacademy-components/src/molecule/learner-skill-card/index.js
```diff
@@ -0,0 +1,169 @@
...
+      color: hovered ? '#FFFFFF' : primarySkinColor,
+      transition: 'background-color 0.15s ease-in-out, color 0.15s ease-in-out'
+    },
+    onExploreClick,
+    'aria-label': buttonExploreAriaLabel ? `${skillTitle}, ${buttonExploreAriaLabel}` : skillTitle,
+    label: buttonExploreLabel,
+    'data-name': 'learner-skill-card-explore-button',
+    icon: {
+      position: 'left',
+      type: 'compass',
```
AdrieanKhisbe, line 61: We might have a little subject here, 

The new designs are supposed to be using fontawesome icons (that were set up in #2837).
But indeed [ButtonLink](https://github.com/CoorpAcademy/components/blob/master/packages/%40coorpacademy-components/src/atom/button-link/index.tsx) is only supporting the old Nova system.

Do you think @HuyIT9898 it's heavy to adapt ButtonLink so it can also support the font awesome icons?
(maybe by considering `icon: {name: ''}` is for fontawesome, while existing `type` while go for Nova  '../../util/button-icons')

If it's too heavy we'll check if compass nova if acceptable

AdrieanKhisbe (commented):
:muscle:

packages/@coorpacademy-components/src/atom/button-link/types.ts
```diff
@@ -33,7 +33,7 @@ export type ButtonLinkProps = {
   'aria-label'?: string;
   'data-name'?: string;
   'data-testid'?: string;
-  icon?: IconType;
+  icon?: IconType & {customStyle?: Record<string, unknown>};
```
adamska27, line 36: why not directly update the Icontype definition above, on line 26?

packages/@coorpacademy-components/src/molecule/learner-skill-card/index.js
```diff
@@ -0,0 +1,169 @@
...
+      customStyle: {
+        height: 16,
+        width: 16
+      }
+    }
+  };
+
+  const completedPercentage = Number.parseInt((completedCourses / skillCourses) * 100);
+
+  const ProgressBar = useCallback(() => {
```
adamska27, line 71: there is already a ProgressBar component, I don't know if you checked it:
src/molecule/progress-bar/index.js

packages/@coorpacademy-components/src/molecule/learner-skill-card/style.css
```diff
@@ -0,0 +1,116 @@
...
+  border-radius: 2px;
+  overflow: hidden;
+  background: light;
+  margin-bottom: 8px;
+}
+
+.progress {
+  display: block;
+  height: 4px;
+  animation: progressbar 2s ease-in-out;
```
adamska27, line 97: progressbar is a reference at?

packages/@coorpacademy-components/src/atom/icon/index.js
HuyIT9898: Ah! I fixed this to `fontSize` because unit test consider it an error and failed

image

packages/@coorpacademy-components/src/molecule/learner-skill-card/style.css
HuyIT9898: Ah, nice catch! I'll update it. thank you Adam

packages/@coorpacademy-components/src/molecule/learner-skill-card/index.js
HuyIT9898: I did check it, but the UI is a bit different, so I'm not sure if we really want to modify and make the original component more complex just for for this use case.

packages/@coorpacademy-components/src/atom/icon/index.js
AdrieanKhisbe: 👌

packages/@coorpacademy-components/src/atom/button-link/types.ts
```diff
@@ -25,7 +25,12 @@ const propTypes = {
 
 export type IconType = {
   position: 'right' | 'left';
-  type: keyof typeof ICONS;
+  type?: keyof typeof ICONS;
+  faIconName?: string;
+  faIconColor?: string;
+  faIconBackgroundColor?: string;
+  faIconSize?: number;
```
youssefezzahi96, line 32: this should help you to avoid using "any" in Button-link component
```suggestion
faIconProps?: PropTypes.InferProps<typeof FaIcon.propTypes>;
```

packages/@coorpacademy-components/src/atom/button-link/index.tsx
```diff
@@ -2,13 +2,46 @@ import React, {useCallback} from 'react';
...
 import classnames from 'classnames';
 import Link from '../link';
+import FaIcon from '../icon';
 import {ICONS} from '../../util/button-icons';
 import propTypes, {ButtonLinkProps, IconType} from './types';
 import style from './style.css';
 
 const getButtonContent = (icon?: IconType, label?: string) => {
-  const {type, position} = icon || {type: '', position: ''};
+  const {type, faIconName, faIconColor, faIconBackgroundColor, faIconSize, position, customStyle} =
```
youssefezzahi96, line 11:
```suggestion
  const {type, faIconProps, position, customStyle} =
```

packages/@coorpacademy-components/src/atom/button-link/index.tsx
```diff
@@ -2,13 +2,46 @@ import React, {useCallback} from 'react';
...
+    icon || {type: '', position: ''};
   const Icon = type && ICONS[type];
+  const FaIconParsed: any = FaIcon;
```
youssefezzahi96, line 14: check my comment in packages/@coorpacademy-components/src/atom/button-link/types.ts: 
using PropTypes.InferProps<typeof FaIcon.propTypes>; you won't have to use any, you can directly do this: 
```suggestion
  if (faIconProps) {
    return (
      <div className={style.buttonContent}>
        {position === 'left' ? <FaIcon {...faIconProps} /> : null}
        {label ? <span className={style.label}>{label}</span> : null}
        {position === 'right' ? <FaIcon {...faIconProps} /> : null}
      </div>
    );
  }
```

packages/@coorpacademy-components/src/atom/button-link/index.tsx
```diff
@@ -2,16 +2,46 @@ import React, {useCallback} from 'react';
...
-  label?: string
-) => {
-  const {type, position, customStyle} = icon || {type: '', position: ''};
+const getButtonContent = (icon?: IconType, label?: string) => {
+  const {type, faIconName, faIconColor, faIconBackgroundColor, faIconSize, position, customStyle} =
+    icon || {type: '', position: ''};
   const Icon = type && ICONS[type];
+  const FaIconParsed: any = FaIcon;
+
+  if (faIconName) {
```
adamska27, line 16: seems like it's the same component here,
I think you can define the component one time before the return, like so:
```
const faIconComponent = (
      <FaIconParsed
        iconName={faIconName}
        iconColor={faIconColor}
        backgroundColor={faIconBackgroundColor}
        size={{
          faSize: faIconSize,
          wrapperSize: faIconSize
        }}
      />
    );
```
and use it after for both case below

packages/@coorpacademy-components/src/atom/icon/index.js
```diff
@@ -62,8 +62,8 @@ const Icon = React.memo(function Icon({
     <div className={style.iconWrapper} style={iconWrapperStyle}>
       <FontAwesomeIcon
         icon={`fa-${iconName}`}
-        color={effectiveIconColor}
-        font-size={effectiveSize.faSize}
+        color={iconColor ?? effectiveIconColor}
```
youssefezzahi96, line 65: you don't need to do this, the line above is already including it 
`const effectiveIconColor =
    iconColor || backgroundColor ? getForegroundColor(backgroundColor) : DEFAULT_ICON_COLOR;`

packages/@coorpacademy-components/src/atom/button-link/index.tsx
AdrieanKhisbe:
```suggestion
  const {type, faIcon, position, customStyle} =
```
`{name, color, backgroundColor, size} = faIcon`

(is customStyle still necessary with the faIcon approach)

packages/@coorpacademy-components/src/atom/button-link/index.tsx
AdrieanKhisbe: :link: https://github.com/CoorpAcademy/components/pull/2846/files#r1574599333

packages/@coorpacademy-components/src/atom/button-link/types.ts
AdrieanKhisbe: :memo: `faIcon`

packages/@coorpacademy-components/src/molecule/learner-skill-card/index.js
```diff
@@ -0,0 +1,170 @@
...
+    );
+  }, [completedPercentage, ariaLabel]);
+
+  return (
+    <div
+      className={style.learnerSkillCardWrapper}
+      data-name="learner-skill-card-wrapper"
+      aria-label={ariaLabel}
+    >
+      <div className={style.skillCoursesAndQuestionsWrapper}>
```
AdrieanKhisbe, line 98: Note, component should support missing kPis (courses, questions, and so on) and hide the section if missing

packages/@coorpacademy-components/src/molecule/learner-skill-card/index.js
```diff
@@ -0,0 +1,170 @@
...
+    'aria-label': ariaLabel,
+    skillCourses,
+    skillQuestions,
+    skillTitle,
+    skillAriaLabel,
+    completedCourses,
+    buttonReviewLabel,
+    buttonReviewAriaLabel,
+    buttonExploreLabel,
+    buttonExploreAriaLabel,
```
AdrieanKhisbe, line 21: I think you should go to a translate approach: `translate('courses')` theses should not be injected from exterior.
We won't need to override them.
(skill we do need are these are going to be dynamic)

packages/@coorpacademy-components/src/molecule/learner-skill-card/index.js
AdrieanKhisbe: (💭  maybe, could be interesting to group all these kpis in a nested objects, like `counts`/`metrics`: `{courses, questions, completedCourses}`)

packages/@coorpacademy-components/src/util/button-icons.js
```diff
@@ -30,6 +31,7 @@ export const ICONS = {
   'chevron-left': ChevronLeftIcon,
   'chevron-right': ChevronRightIcon,
   close: CloseIcon,
+  compass: CompassIcon,
```
AdrieanKhisbe, line 34: Maybe you can revert it with the new fa approach you set up

AdrieanKhisbe (changes requested):

packages/@coorpacademy-components/src/atom/icon/index.js
HuyIT9898: without the modify, the outcome color is not as expected for both normal and hover state

image


image

packages/@coorpacademy-components/src/atom/button-link/types.ts
HuyIT9898: I got this error when tried the above code:

image

I think the reason is that Icon is wrap with React.Memo, so it only have type of React.Memo return value `React.NamedExoticComponent<object>` which typescript is unable to know the inner props.
This is the same reason that I have to use any to parsed the Icon component to type any to passin props without error (I've updated the code to use other method to pass in the prop without using the "any" type)

packages/@coorpacademy-components/src/atom/button-link/index.tsx
HuyIT9898: Thank you @youssefezzahi96 @AdrieanKhisbe , I've updated

packages/@coorpacademy-components/src/molecule/learner-skill-card/index.js
HuyIT9898: Thank you, I've updated

packages/@coorpacademy-components/src/molecule/learner-skill-card/index.js
HuyIT9898: thank you, I've updated

packages/@coorpacademy-components/src/atom/button-link/index.tsx
```diff
@@ -2,14 +2,45 @@ import React, {useCallback} from 'react';
...
+        iconName: name,
+        iconColor: color,
+        backgroundColor,
+        size: {
+          faSize: size,
+          wrapperSize: size
+        }
+      }}
+    />
+  );
```
AdrieanKhisbe, line 32: this and descructuing should be in the `if (faIcon) {` branch!

```js
if (faIcon) {
 const {name, color, backgroundColor, size} = faIcon
 const faIconComponent = (...)
 return (<div>....)
}
```

related: what is default behavior for `position`? :`left` (according `type` is must be one of these)
(or absence implies the icon does not appear?)

AdrieanKhisbe: > (just an idea, not to apply in practice to keep same structure than last return)
> 
> about the diff, alternative can be to have to two element array, and reverse it if right position
> 
> ```js
> const buttonParts = [ faIconComponent /* could be inlined*/, label ? <span className={style.label}>{label}</span> : null]
> return (
>     <div className={style.buttonContent}>
>         {position === 'right' ? _.reverse(buttonParts) : buttonParts}
>     </div>
> )
> ```
> (`reverse` from `lodash/fp`)

packages/@coorpacademy-components/src/molecule/learner-skill-card/index.js
```diff
@@ -0,0 +1,180 @@
...
+  const {skin, translate} = context;
+  const [hovered, setHovered] = useState(false);
+  const primarySkinColor = getOr('#0061FF', 'common.primary', skin);
+
+  const review = translate('Review');
+  const explore = translate('Explore');
+  const courses = translate('courses');
+  const questions = translate('questions');
+  const skillFocus = translate('skill_focus');
+  const coursesCompleted = translate('courses_completed');
```
AdrieanKhisbe, line 29: (I might rename with Locale suffix to disambiguitate in the component building around line 100 (ex [line 104](https://github.com/CoorpAcademy/components/pull/2846/files#diff-7a6de17128fffd61ef831601136afb36293e13abca97f66b1342f2a7972d047aR104)

packages/@coorpacademy-components/src/molecule/learner-skill-card/index.js
```diff
@@ -0,0 +1,180 @@
...
+  const ProgressBar = useCallback(() => {
+    const progressBarColor = '#3EC483';
+    const inlineProgressValueStyle = {
+      backgroundColor: progressBarColor,
+      width: `${completedPercentage}%`
+    };
+
+    if (!skillCourses) {
+      return null;
+    }
```
AdrieanKhisbe, line 79: you can exit earlier, and one line it  `if (!skillCourses) return null`.

Or even set the ProgressBar itself to null.  (I'm unsure about that part, I'm rusty with react component lifecycle)

packages/@coorpacademy-components/src/molecule/learner-skill-card/test/fixtures/default.js
```diff
@@ -0,0 +1,16 @@
+export const defaultProps = {
+  'aria-label': 'Learner Skill Card',
+  skillTitle: 'Critical thinking and analysis',
+  skillAriaLabel: 'Skill aria label',
+  metrics: {
```
AdrieanKhisbe, line 5: 👍

packages/@coorpacademy-components/src/atom/button-link/index.tsx
AdrieanKhisbe: Suggestion for rework of the whole function in three step
1. early exist for no icon case
2. forge icon component (depending on icon or faicon)
3. return with proper order of icon and label

```
if (!Icon && !faIcon)  return no icon component

const iconComponent  = faIcon ? <FaIcon iconNam={faIcon.name} ... /> : <Icon className={style.icon} theme="coorpmanager" />
return (
    <div className={style.buttonContent}>
      {position === 'left' ? iconComponent : null}
      {label ? <span className={style.label}>{label}</span> : null}
      {position === 'right' ? iconComponent : null}
    </div>
  );
```

packages/@coorpacademy-components/src/atom/button-link/index.tsx
```diff
@@ -9,15 +9,24 @@ import style from './style.css';
...
     backGroundColor: '',
     size: 0
   };
-  const Icon = type && ICONS[type];
 
-  const faIconComponent = (
+  const iconComponent = faIcon ? (
     <FaIcon
       {...{
         iconName: name,
```
AdrieanKhisbe, line 32: using `faIcon.name` and so on, you wont need destructuring faIcon and fallbacl on line before

AdrieanKhisbe (commented):
Just a last follow up, and it's good :)

packages/@coorpacademy-components/src/atom/button-link/index.tsx
AdrieanKhisbe: (in other words: remove `const {name, color, backgroundColor, size} = faIcon || {` and `iconName: faIcon.name, color: faIcon.color, ...`
(fallback might not be necessary)

adamska27 (approved):

AdrieanKhisbe (approved):

packages/@coorpacademy-components/src/atom/icon/index.js
youssefezzahi96: @HuyIT9898 
this should resolve your problem, yeah just parentheses :p 
`
  const effectiveIconColor =
    iconColor || (backgroundColor ? getForegroundColor(backgroundColor) : DEFAULT_ICON_COLOR);
`

youssefezzahi96 (changes requested):

youssefezzahi96 (approved):

PR #2867, opened 2024-07-15
CLXP-57 + CLXP-80 Skill detail template

HuyIT9898:
Purpose: Skill detail template, Discipline skill tag hover effect

review app: https://6628d2e0fd4aa63f3be2e1ab-tzvrqblqyh.chromatic.com/?path=/story/template-skilldetail--default

PR #2870, opened 2024-08-06
Fix skillBaseUrl not passed properly in Discipline template

HuyIT9898:
Purpose: Fix skillBaseUrl not passed properly in Discipline template

Fix #2867


AdrieanKhisbe (approved):
Please refer next time the PR that introduce the issue.
#2867 in this case if i'm not wrong

PR #2869, opened 2024-08-02
pass disabled props as button attribute

adamska27:
https://go1web.atlassian.net/jira/software/projects/CLXP/boards/626?selectedIssue=CLXP-86

**Detailed purpose of the PR**

fix disabled behavior on following components:
* buttonLink
* butonLinkOnly
* buttonMenu
* Link (and fix fixtures)

**Result and observation**

- [ ] **Breaking changes ?**  
       If checked, what have you broken ?

- [ ] **Extra lib ?**
      If checked, Which extra lib did you add ? (name, purpose, link ...).

**Testing Strategy**

- [x] Already covered by tests
- [x] Manual testing
- [ ] Unit testing

PR #2871, opened 2024-08-06
Link href fix

adamska27:
previous PR which change the Link component behavior: https://github.com/CoorpAcademy/components/pull/2869

explain why PR update dependencies fail: https://github.com/CoorpAcademy/coorpacademy/pull/18117

- restore Link component behavior when there is no href props -> no href attribute
- better disabled support: when disabled -> not action on onClick 

**Result and observation**

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [ ] Unit testing

### DIFF
--- a/packages/@coorpacademy-components/src/atom/link/index.js
+++ b/packages/@coorpacademy-components/src/atom/link/index.js
@@ -74,8 +74,8 @@ const Link = (props, legacyContext) => {
   );
 
   const _href = useMemo(() => {
-    if (!href) return;
     if (disabled) return '#';
+    if (!href) return;
 
     return createHref(href);
   }, [createHref, disabled, href]);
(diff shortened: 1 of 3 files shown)


packages/@coorpacademy-components/src/atom/link/index.js
```diff
@@ -72,13 +73,20 @@ const Link = (props, legacyContext) => {
     [hoverColor, hovered, skinHover]
   );
 
+  const _href = useMemo(() => {
+    if (disabled) return '#';
```
HuyIT9898, line 77: Maybe we can merge this and the line below into `if (disabled || !href) return;`?

adamska27: yes, I agree, I changed that

HuyIT9898 (approved):

AdrieanKhisbe (approved):

PR #2872, opened 2024-08-07
CLXP-110 Skill picker modal footer is hidden when zoom in on browser

HuyIT9898:
Purpose: fix Skill picker modal footer is hidden when zoom in on browser

review app: https://6628d2e0fd4aa63f3be2e1ab-ywbzvisqli.chromatic.com/

125%: 
image

150%:

image


AdrieanKhisbe (approved):
:+1:

PR #2876, opened 2024-08-20
CLXP-114 Enhancement Button Link component to have it own hover state

HuyIT9898:
Purpose: Enhancement Button Link component to have it own hover state

ticket: https://go1web.atlassian.net/browse/CLXP-114

review app: https://6628d2e0fd4aa63f3be2e1ab-kubwsfxkvp.chromatic.com/

### DIFF
--- a/packages/@coorpacademy-components/src/atom/button-link/index.tsx
+++ b/packages/@coorpacademy-components/src/atom/button-link/index.tsx
@@ -1,4 +1,4 @@
-import React, {useCallback} from 'react';
+import React, {useCallback, useState, useMemo} from 'react';
 import {noop} from 'lodash/fp';
 import classnames from 'classnames';
 import Link from '../link';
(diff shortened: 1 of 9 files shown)


packages/@coorpacademy-components/src/molecule/learner-skill-card/index.js
```diff
@@ -49,10 +44,12 @@ const LearnerSkillCard = (props, context) => {
 
   const buttonExploreProps = {
```
adamska27, line 45: can you do the same thing for the learning-priority-card? It uses the same old logic

I merged the pull request, so you maybe need to rebase,

thanks

packages/@coorpacademy-components/src/atom/button-link/index.tsx
```diff
@@ -74,10 +82,28 @@ const ButtonLink = (props: ButtonLinkProps) => {
...
+            backgroundColor: hoverBackgroundColor,
+            color: hoverColor
+          }
+        : null)
+    };
+  }, [hoverBackgroundColor, hoverColor, hovered, customStyle]);
+
   if (link) {
     return (
       <Link
```
adamska27, line 109: maybe, it' a good idea to enable the same logic for the Link component too ?

packages/@coorpacademy-components/src/molecule/learner-skill-card/index.js
HuyIT9898: Thanks Adem, I've updated for it too!

packages/@coorpacademy-components/src/atom/button-link/index.tsx
HuyIT9898: Ah, the Link itself already have hover state logic: https://github.com/CoorpAcademy/components/blob/master/packages/%40coorpacademy-components/src/atom/link/index.js#L9

adamska27 (approved):

PR #2893, opened 2024-10-29
CLXP-127 Technical: Missing translation shouldn't break the UX on CM

HuyIT9898:
Add a flag `isAllowFalsy` to `createTranslate` so that error is not throw if locale key is missing if  `isAllowFalsy === true`

**Testing Strategy**

- [x] Manual testing
- [x] Unit testing

### DIFF
--- a/packages/@coorpacademy-translate/src/test/translate.js
+++ b/packages/@coorpacademy-translate/src/test/translate.js
@@ -118,7 +118,7 @@ test("should throw error if nested template couldn't be found", t => {
   t.throws(() => translate('foo.foo_none'), {message: 'Key foo.foo_none not found!'});
 });
 
-test("should not throw error if template couldn't be found and isAllowFalsy === true", t => {
+test("should not throw error if template couldn't be found and throwIfMissing === true", t => {
   const translate = createTranslate(
     {
       foo: 'Foo'
--- a/packages/@coorpacademy-translate/src/test/translate.js
+++ b/packages/@coorpacademy-translate/src/test/translate.js
@@ -119,9 +119,12 @@ test("should throw error if nested template couldn't be found", t => {
 });
 
 test("should not throw error if template couldn't be found and isAllowFalsy === true", t => {
-  const translate = createTranslate({
-    foo: 'Foo'
-  }, true);
+  const translate = createTranslate(
+    {
+      foo: 'Foo'
+    },
+    true
+  );
 
   t.is(translate('bar'), 'bar');
 });
--- a/packages/@coorpacademy-translate/src/test/translate.js
+++ b/packages/@coorpacademy-translate/src/test/translate.js
@@ -117,3 +117,11 @@ test("should throw error if nested template couldn't be found", t => {
 
   t.throws(() => translate('foo.foo_none'), {message: 'Key foo.foo_none not found!'});
 });
+
+test("should not throw error if template couldn't be found and isAllowFalsy === true", t => {
+  const translate = createTranslate({
+    foo: 'Foo'
+  }, true);
+
+  t.is(translate('bar'), 'bar');
+});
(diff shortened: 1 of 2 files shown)


packages/@coorpacademy-translate/src/translate.js
```diff
@@ -16,21 +16,25 @@ function getTemplate(locales, key, count) {
   return getOr(regularTemplate, `${key}_plural`, locales);
 }
 
-const createTranslate = locales => (key, data) => {
-  const template = getTemplate(locales, key, get('count', data));
-  if (!isString(template)) {
-    throw new Error(`Key ${key} not found!`);
-  }
+const createTranslate =
+  (locales, isAllowFalsy = false) =>
```
AdrieanKhisbe, line 20: I'd give a more functional name to the arg, like `throwIfMissing` (which invert the values)
```suggestion
  (locales, throwIfMissing = true) =>
```

packages/@coorpacademy-translate/src/translate.js
```diff
@@ -16,21 +16,25 @@ function getTemplate(locales, key, count) {
...
-  return replace(
-    interpolation,
-    (token, value) => {
-      const _value = trim(value);
-      return has(_value, data) ? get(_value, data) : token;
-    },
-    template
-  );
-};
+    return !isString(template)
```
AdrieanKhisbe, line 27: tip: if you want to minimize change you can just `if (!isString(template) return key`

AdrieanKhisbe: if fact, you can simplify the implementation. 

Pseudocode:
```
 if !isString(template):
   if throwIfMissing: throw
   return key

 return replace 
```

AdrieanKhisbe (changes requested):
Few implementation feedbacks/suggestions.
Also, could you mention the new option on the package README?

packages/@coorpacademy-translate/src/translate.js
```diff
@@ -16,21 +16,24 @@ function getTemplate(locales, key, count) {
   return getOr(regularTemplate, `${key}_plural`, locales);
 }
 
-const createTranslate = locales => (key, data) => {
-  const template = getTemplate(locales, key, get('count', data));
-  if (!isString(template)) {
-    throw new Error(`Key ${key} not found!`);
-  }
+const createTranslate =
+  (locales, throwIfMissing = false) =>
```
AdrieanKhisbe, line 20: You want to preserve the existing behavior as default
```suggestion
  (locales, throwIfMissing = true) =>
```

AdrieanKhisbe (changes requested):
oupsi, I didnt know there was no README 🙈 

https://github.com/CoorpAcademy/components/tree/master/packages/%40coorpacademy-translate

packages/@coorpacademy-translate/src/translate.js
```diff
@@ -16,21 +16,24 @@ function getTemplate(locales, key, count) {
...
+  (key, data) => {
+    const template = getTemplate(locales, key, get('count', data));
+    if (!isString(template)) {
+      if (!throwIfMissing) throw new Error(`Key ${key} not found!`);
```
AdrieanKhisbe, line 24:
```suggestion
      if (throwIfMissing) throw new Error(`Key ${key} not found!`);
```

packages/@coorpacademy-translate/src/test/translate.js
```diff
@@ -117,3 +117,14 @@ test("should throw error if nested template couldn't be found", t => {
 
   t.throws(() => translate('foo.foo_none'), {message: 'Key foo.foo_none not found!'});
 });
+
+test("should not throw error if template couldn't be found and throwIfMissing === true", t => {
+  const translate = createTranslate(
+    {
+      foo: 'Foo'
+    },
+    true
```
AdrieanKhisbe, line 126:
```suggestion
    false
```

AdrieanKhisbe (changes requested):
(complement to previous one)

AdrieanKhisbe (approved):

PR #2902, opened 2024-12-04
CLXP-192 Update Skill detail page progress bar section

HuyIT9898:
→ Delete the Your progress section

→ Move the “X question to review” below the description

If X = 0 then the Question to review are not displayed

→ Move the progress bar just below the description

review app: https://6628d2e0fd4aa63f3be2e1ab-hyvpsypsij.chromatic.com/?path=/story/template-skilldetail--default

**Testing Strategy**

- [ ] Manual testing
- [ ] Unit testing


adamska27:
image

There is an issue on the AllCourses component, for the skill details page specifically, the number for All content is not correct, 
for the skills with use pagination, so we have 32 content instead of 84 for example,

maybe, we have to add a property totalContent for this specific case

AdrieanKhisbe (commented):
While filtering based on a search term, clicking on one of the content types button make the filter disappear.
Is it a glitch, or is this because of the storybook preview?

PR #2909, opened 2025-01-07
Cplx 192 2

adamska27:

global:
- add missing locales

all certification page:
- add missing radius
- increase img badge height to better display rectangular img

certification details
- set a max width to avoid badge img to overflow
- fix show more display

**Result and observation**

before radius:
image

after radius:
image

before max width:
image

before img badge rectangular:
image

after img badge rectangular:
image

after max width:
image

before no show more because display show more based on length:
image

after show more because based on element height:
image

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [ ] Unit testing


packages/@coorpacademy-components/locales/en/global.json
```diff
@@ -42,6 +43,7 @@
   "continue_learning": "Continue learning",
   "learning_priority_modal_title": "New learning priority",
   "learning_priority_modal_description": "Create a learning priority by selecting either one skill, playlist or certification.",
+  "locked": "locked",
```
silouone, line 46: L maj?

adamska27: ah mince

silouone (approved):

PR #3018, opened 2025-09-08
Add quick filters into search page

esa-coorp:
Capture d’écran 2025-09-08 à 10 12 00

- [ ] **Breaking changes ?**  

- [ ] **Extra lib ?**

**Testing Strategy**

- [ ] Manual testing
- [ ] Unit testing


packages/@coorpacademy-components/src/template/common/search-page/index.js
```diff
@@ -134,7 +136,9 @@ const SearchPage = (props, context) => {
           filterGroupAriaLabel={filterGroupAriaLabel}
           sortAriaLabel={sortAriaLabel}
         />
-      ) : null}
+      ) : (
+        <QuickFilters {...quickFilters} />
```
DjamelSoualmi, line 140: sans condition sur quickFilters ?

PR #3031, opened 2026-06-05
[codex] fix aria prohibited attributes

esa-coorp:
## What changed

- Adds compatible roles to components that expose `aria-label` on generic wrappers, including CSS-backed images, grouped containers, and custom controls.
- Moves the custom select accessible label from the visual wrapper to the actual button.
- Adds keyboard activation for clickable `div` controls such as QCM answers, chips, search reset, and learner skill cards.
- Removes literal dashboard banner labels such as `{title}` and `{subtitle}` where visible text already provides the accessible name.

## Why

Axe reports `aria-prohibited-attr` when ARIA naming attributes are used on elements whose role does not permit them. Several components were applying `aria-label` to generic `div`/`span` wrappers, which can trigger WCAG 4.1.2 failures.

### DIFF
--- a/packages/@coorpacademy-components/src/atom/chip/index.js
+++ b/packages/@coorpacademy-components/src/atom/chip/index.js
@@ -17,6 +17,14 @@ const LUMINOSITY_DELTA = 0.08;
 const {cm_primary_blue: DEFAULT_BACKGROUND_COLOR} = COLORS;
 const ICON_SIZE = '12px';
 
+const handleChipKeyDown = onClick => event => {
+  if (!onClick) return;
+  if (event.key === 'Enter' || event.key === ' ') {
+    event.preventDefault();
+    onClick(event);
+  }
+};
+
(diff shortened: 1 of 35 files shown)


silouone (approved):

adamska27 (approved):

esa-coorp:
CI note from the Travis log: the failing job exits during `npx lerna run lint` because `@coorpacademy/history:lint` ends with `Segmentation fault`.

The `@coorpacademy/components:lint` task is listed under “Tasks not run because their dependencies failed or --nx-bail=true”, so this failure does not point to the ARIA changes in this PR. Local validation for the touched package still passes:

- targeted AVA tests: 12 passed
- static ARIA scan: `count=0`
- `npm run typecheck`: OK
- eslint on changed files: no errors

This looks like a Travis/Nx lint flake or infra issue in `@coorpacademy/history`; rerunning the Travis job is the appropriate next step.

